PR #1, opened 2015-07-16
Initial dynflow infrastructure for the smart proxy

iNecas:


jlsherrill:
Would be good to have rubocop in from the start

lib/smart_proxy_dynflow.rb
```diff
@@ -0,0 +1,66 @@
...
+      yield config if block_given?
+    end
+  end
+
+  def logger_adapter
+    ::Dynflow::LoggerAdapters::Simple.new $stderr, 0
+  end
+
+  def web_console
+    require 'dynflow/web_console'
```
jlsherrill: isn't this deprecated? :)

https://github.com/Dynflow/dynflow/blob/master/lib/dynflow/web_console.rb#L3

iNecas: good poing :) I should be aware of the deprecations I've introduced myself

iNecas: Updated

iNecas:
Rubocop added

smart_proxy_dynflow.gemspec
```diff
@@ -0,0 +1,34 @@
...
+    Use the Dynflow inside Foreman smart proxy
+  EOS
+
+  gem.files         = Dir['{bundler.d,lib,settings.d}/**/*', 'LICENSE', 'Gemfile']
+  gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
+  gem.require_paths = ["lib"]
+  gem.license = 'GPLv3'
+
+  gem.add_development_dependency "bundler", "~> 1.7"
+  gem.add_development_dependency "rake", "~> 10.0"
```
stbenjam: Two deps for rake in the Gemspec. https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/1/files#r36130965

settings.d/dynflow.yml.example
```diff
@@ -0,0 +1,2 @@
+---
+:enabled: false
```
stbenjam: Is there a reason to be false by default? One generally wants by installing the package for it to be on by default, unless there's additional configuration.

The RPM will copy the example to the real yaml.

ares: If there are no other manual options we can make it enabled by default, btw user will have to restart the proxy manually anyway. OTOH if other plugins are disabled by default, I'd keep it same for consistency. IIRC the installer usually configures these files so this case applies only when it's installed manually from package.

iNecas: Originally, I took the smart_proxy_chef as the scafold for the plugin structure https://github.com/theforeman/smart_proxy_chef/blob/master/settings.d/chef.yml.example, also, all the other features in smart proxy are disabled by default https://github.com/theforeman/smart-proxy/tree/develop/config/settings.d (although the main reason for this was probably because the features are not separated into rpms), I see the majority of other plugins are enabled by default, so I will do the same here

iNecas: Updated

smart_proxy_dynflow.gemspec
iNecas: Fixed

iNecas:
Merging and going to relese 0.0.1 version :)

PR #4, opened 2015-08-14
Postpone the Dynflow world initialization

iNecas:
We can't initalize the dynflow world sooner than the smart proxy forks in
production, because otherwise we loose the file descriptors we need to open at
the initialization phase.

We also add a hooking mechanism for other plugins using dynflow/concurrent
ruby for after_initialization tasks


stbenjam:
ACK

iNecas:
Thanks

PR #7, opened 2015-09-15
refs #11827 - add api call to return count of tasks with a state

stbenjam:
I want to be able to query the proxy for the number of currently running actions before distributing the new tasks when doing the load balancing.

It's a pretty simplistic round robin, not sure if want to do something fancier.

### DIFF
--- a/lib/smart_proxy_dynflow/helpers.rb
+++ b/lib/smart_proxy_dynflow/helpers.rb
@@ -23,6 +23,12 @@ module Proxy
         status 404
         {}
       end
+
+      def tasks_count(state)
+        filter = state ? { :filters => { :state => [state] } } : {}
+        tasks = world.persistence.find_execution_plans(filter)
+        { :count => tasks.count, :state => state || 'all' }
+      end
     end
   end
 end
--- a/lib/smart_proxy_dynflow/api.rb
+++ b/lib/smart_proxy_dynflow/api.rb
@@ -23,6 +23,10 @@ module Proxy
       get "/tasks/:task_id/status" do |task_id|
         task_status(task_id).to_json
       end
+
+      get "/tasks/count" do
+        tasks_count(params['state']).to_json
+      end
     end
   end
 end
--- a/test/api_test.rb
+++ b/test/api_test.rb
@@ -62,5 +62,20 @@ class Proxy::Dynflow
         execution_plan.result.must_equal :success
       end
     end
+
+    describe 'GET /tasks/count' do
+      it 'counts the actions in state' do
+        get "/tasks/count", :state => 'stopped'
+        response = JSON.load(last_response.body)
+        old_count = response['count']
+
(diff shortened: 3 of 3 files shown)


jlsherrill:
Code looks fine and works for me.  ACK from my side.  Will let @iNecas decide if he likes the concept

iNecas:
Let's wait for the foreman server part: so far it looks good, just in case we would need some additional change on this API that are not abviour right now

lib/smart_proxy_dynflow/helpers.rb
```diff
@@ -23,6 +23,11 @@ def task_status(task_id)
         status 404
         {}
       end
+
+      def tasks_count(state)
+        tasks = world.persistence.find_execution_plans(:filters => {:state => [state]})
```
adamruzicka: This would make rubocop happier `{ :state => [state] }`

adamruzicka:
Wouldn't it better to change the route to `get /tasks/by-status/:state` to keep the `get /tasks/:task_id` free for retrieving info about a certain task or something?

iNecas:
Good point `/tasks/state`, with accepting the state in params, woudl work as well

stbenjam:
Updated - how about get `/tasks/count?state=running`?

iNecas:
works for me

lib/smart_proxy_dynflow/helpers.rb
```diff
@@ -23,6 +23,12 @@ def task_status(task_id)
         status 404
         {}
       end
+
+      def tasks_count(state)
+        filter = state ? { :filters => { :state => [state] } } : {}
+        tasks = world.persistence.find_execution_plans(filter)
+        { :count => tasks.count, :state => state || 'all' }
```
adamruzicka: `tasks_count(nil)` would return
```ruby
{ :count => 1, :state => 'all' }
```

but `tasks_cound('all')` would return
```ruby
{ :count => 0, :state => 'all' }

**Comment by adamruzicka** [comment]
Left one inline comment, except for that looks good to me, ACK

File: lib/smart_proxy_dynflow/helpers.rb (hunk 2 above)
Comment (iNecas) [inline lib/smart_proxy_dynflow/helpers.rb]: This shoud be enough

```ruby
        state ||= 'all'
        filter = state != 'all' ? { :filters => { :state => [state] } } : {}
```

adamruzicka: +1

iNecas: I've updated the code before merging, so that we can proceed with PRs on foreman-tasks side

iNecas:
Meged as https://github.com/theforeman/smart_proxy_dynflow/commit/527d0f2e2b7d8f3f066ee8269132afbebc3c333e, thanks @stbenjam

PR #9, opened 2016-01-19
Fixes #11366 - adds dynflow persistance

ares:
will need to create `/var/spool/foreman-proxy/dynflow` in package spec, see https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-smart_proxy_openscap/rubygem-smart_proxy_openscap.spec#L5 for an example


iNecas:
Tested and works great! Thanks @ares

PR #10, opened 2016-02-02
Fixes #13523 - use in-memory sqlite for tests

iNecas:
No need to worry about the creation of the sqlite file then.


stbenjam:
ACK

iNecas:
thanks @stbenjam

PR #11, opened 2016-02-04
fixes #13564 - use /var/lib instead of /var/spool

stbenjam:


adamruzicka:
ACK, `/var/lib/` looks like a better place for it

PR #12, opened 2016-02-15
Fixes #11363 - Smart proxy dynflow console auth

adamruzicka:
Plain HTTP remains freely accessible, connections over HTTPS require a client SSL certificate

### DIFF
--- a/settings.d/dynflow.yml.example
+++ b/settings.d/dynflow.yml.example
@@ -1,3 +1,4 @@
 ---
 :enabled: true
 :database: /var/lib/foreman-proxy/dynflow/dynflow.sqlite
+:console_auth: true
--- a/lib/smart_proxy_dynflow.rb
+++ b/lib/smart_proxy_dynflow.rb
@@ -65,6 +65,12 @@ class Proxy::Dynflow
         # closing opened file descriptors)
         # TODO: extend smart proxy to enable hooks that happen after
         # the forking
+
+        if Proxy::Dynflow::Plugin.settings.console_auth
+          authorize_with_trusted_hosts
+          authorize_with_ssl_client
+        end
+
         Proxy::Dynflow.ensure_initialized
         set :world, Proxy::Dynflow.world
       end
--- a/lib/smart_proxy_dynflow/plugin.rb
+++ b/lib/smart_proxy_dynflow/plugin.rb
@@ -5,6 +5,7 @@ class Proxy::Dynflow
 
     settings_file "dynflow.yml"
     default_settings :database => '/var/lib/foreman-proxy/dynflow/dynflow.sqlite'
+    default_settings :console_auth => true
     plugin :dynflow, Proxy::Dynflow::VERSION
   end
 end


iNecas:
Thinking: maybe we should make the option configurable + the trusted host check (default to false, but useful for the http users).

settings.d/dynflow.yml.example
```diff
@@ -1,3 +1,5 @@
 ---
 :enabled: true
 :database: /var/lib/foreman-proxy/dynflow/dynflow.sqlite
+:dynflow_console_authorize_with_ssl_client: false
```
iNecas: I would go with true by default for `dynflow_console_authorize_with_ssl_client`

iNecas: Also, since it's a dynflow plugin, we can assume, that the console is dynflow as well, so we can shorten the option to `console_authorize_with_ssl_client`

settings.d/dynflow.yml.example
```diff
@@ -1,3 +1,5 @@
 ---
 :enabled: true
 :database: /var/lib/foreman-proxy/dynflow/dynflow.sqlite
+:console_authorize_with_ssl_client: true
+:console_authorize_with_trusted_hosts: false
```
stbenjam: I'd prefer if these were shorter, as they will need to be installer options as well.

Maybe console_ssl and console_trusted_hosts instead.

adamruzicka: Shorter names would be nice, but at the cost of being less self-explanatory. How about something like `console_ssl_auth` and `console_trusted_hosts_auth`?

stbenjam: This gets resolved to something like `--foreman-proxy-dynflow-console-trusted-hosts-auth`  or similar, it's still very long.

What about just making this one setting? Say, `console_auth` which would turn off both trusted_hosts and ssl.   I would guess the only reason to do this is in a situation where we want to collect some debug information and have the user directly access the console. Right?

In every other situation, I think we should let the smart proxy handle it according to its own settings.

adamruzicka: Hmm, that's really long. Bundling both into one option feels wrong, considering those are two different access control mechanisms.

@ares, @iNecas: any opinions on this?

stbenjam: IMHO it feels wrong for us to touch these at all.  It a user has configured trusted_hosts or ssl client certificates in settings.yml, they should assume it protects all of the sensitive components.  The dynflow console being one of them.

Why are we creating a second set of options @iNecas? I guessed it was to make it easier to get into the console if a troubleshooting case, and in that case, you would be concerned with turning both off.

iNecas: Yes. First we thought that we don't want to make the `trusted_hosts` check turned on by default, as the admin could access the host from differen host, than the foreman one, on the other hand, the proxy uses just the cn in the cert (and a user could even add their CN into the trusted hosts if needed), which is better, than trusting anyone with a valid cert (even though just puppet client). So +1 for just on/off for authorizing the console access and using both trusted host and certs mechanism

stbenjam:
Great, thanks! ACK

iNecas:
Thanks @adamruzicka and @stbenjam

PR #14, opened 2016-04-04
Fixes #14460 - Add ability to use local Gemfiles

cpeters:


stbenjam:
Oops, sorry, looks like we did turn on redmine issues for the repo.

@cpeters Can you file one please?

cpeters:
TEST

cpeters:
@stbenjam Updated

stbenjam:
Shouldn't we do the same as Foreman and Smart Proxy and read a `Gemfile.local.rb` it from bundler.d? Or is there somewhere else that the other repos support this Gemfile.local thing that I missed?

They seem to do it this way:

```
Dir["#{File.dirname(__FILE__)}/bundler.d/*.rb"].each do |bundle|
  self.instance_eval(Bundler.read_file(bundle))
end
```

cpeters:
@stbenjam The hammer cli plugins do it with a Gemfile.local file.

https://github.com/theforeman/hammer-cli-foreman-discovery/blob/master/Gemfile#L18-L19

I'm not opposed to doing it with bundler.d similar to Foreman

stbenjam:
Ok cool, how about Foreman's style then?

shlomizadok:
+1 for Foreman style

cpeters:
@stbenjam @shlomizadok Updated

stbenjam:
Oh, `bundler.d` is already used, the dynflow.rb is copied to the smart proxy dir using this path.  It's the same in all the smart proxy plugins.

See packaging: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec#L63

So I'm not sure, maybe we should adopt the hammer style.

cpeters:
Yeah. I tried modifying the gemspec, removing the bundler.d from gem.files and still ran into a bundler issue.

stbenjam:
The packaging gets it from the gem, so it needs to stay.

cpeters:
@stbenjam Ok; back to the original with pulling Gemfile.local into the Gemfile.

stbenjam:
Looks good to me, thanks! Will merge tomorrow, just want to give the others on @theforeman/remote-execution a chance to make sure it's ok for them.

Gemfile
```diff
@@ -6,3 +6,7 @@ group :development do
   gem 'smart_proxy', :git => "https://github.com/theforeman/smart-proxy", :branch => "develop"
   gem 'pry'
 end
+
+# load local gemfile
+local_gemfile = File.join(File.dirname(__FILE__), 'Gemfile.local')
```
iNecas: bikesheding: `Gemfile.local.rb` would be better, as editors recognize it's a ruby source.

Also adding it to `.gitignore` would make sense here

iNecas:
Thanks @stbenjam for waiting :)

cpeters:
@iNecas Updated

iNecas:
:+1:

stbenjam:
I was inclined to also make the `.rb` comment, but didn't because we'll have 3 different ways of doing this in the Foreman org (bundler.d, Gemfile.local.rb, and Gemfile.local).  :crying_cat_face:  https://xkcd.com/927/

Anyway, it's the right solution.  I opened a RM against hammer anyway. http://projects.theforeman.org/issues/14466

stbenjam:
Merging, thanks @cpeters, @iNecas

PR #15, opened 2016-04-21
Fixes #14901 - Api part

adamruzicka:
API part of taking dynflow out of the smart proxy to separate process, which lives [here](https://github.com/adamruzicka/smart_proxy_dynflow/tree/core) until I figure out a proper name for it


lib/smart_proxy_dynflow/plugin.rb
```diff
@@ -1,11 +1,14 @@
...
+require 'proxy/plugin'
 class Proxy::Dynflow
   class Plugin < Proxy::Plugin
     http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
     https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
 
     settings_file "dynflow.yml"
-    default_settings :database => '/var/lib/foreman-proxy/dynflow/dynflow.sqlite'
     default_settings :console_auth => true
+    default_settings :core_url => 'http://localhost:8001'
```
stbenjam: I know this is a work in progress, but had an initial thought, it'd be more convenient if the service would listen on a port marked already for HTTP so we don't need to touch SELinux.  These are the options:

```
http_port_t                    tcp      80, 81, 443, 488, 8008, 8009, 8443, 9000
```

I think 8008 is free on a proxy/capsule.

adamruzicka: Sounds good, I just took first port at hand

config/settings.yml.example
```diff
@@ -0,0 +1,10 @@
+---
```
ares: I think these options deserves some comments explaining their meaning, also I wonder why capital letters for Host and Port? I'd like to keep it consistent if possible

smart_proxy_dynflow_core.gemspec
```diff
@@ -0,0 +1,37 @@
...
+  gem.authors       = ['Ivan Nečas']
+  gem.email         = ['inecas@redhat.com']
+  gem.homepage      = "https://github.com/theforeman/smart_proxy_dynflow"
+  gem.summary       = "Dynflow runtime for Foreman smart proxy"
+  gem.description   = <<-EOS
+    Use the Dynflow inside Foreman smart proxy
+  EOS
+
+  gem.files         = Dir['lib/smart_proxy_dynflow_core.rb', 'config/settings.yml.example',
+                          '{bundler.d, lib/smart_proxy_dynflow_core/*}', 'LICENSE', 'Gemfile']
```
ares: missing a bin script to start core

ares: missing `**` in `lib/smart_proxy_dynflow_core/*` so only one file is actually included

adamruzicka: The problem here is not missing `**`, but having the `/*` inside the curly braces. This should be fixed in https://github.com/theforeman/smart_proxy_dynflow/pull/15/commits/c01a35f4d9b823184ea28b21e1a7fc552bad1ce4 . Will add the script from bin

Gemfile
```diff
@@ -1,12 +1,14 @@
...
 
-gemspec
+gemspec :name => 'smart_proxy_dynflow_core'
 
 group :development do
   gem 'smart_proxy', :git => "https://github.com/theforeman/smart-proxy", :branch => "develop"
   gem 'pry'
 end
 
+gem 'smart_proxy_dynflow', :path => '.'
```
ares: why this :path => '.' is needed?

smart_proxy_dynflow_core.gemspec
```diff
@@ -0,0 +1,38 @@
...
+  gem.authors       = ['Ivan Nečas']
+  gem.email         = ['inecas@redhat.com']
+  gem.homepage      = "https://github.com/theforeman/smart_proxy_dynflow"
+  gem.summary       = "Dynflow runtime for Foreman smart proxy"
+  gem.description   = <<-EOS
+    Use the Dynflow inside Foreman smart proxy
+  EOS
+
+  gem.files         = Dir['lib/smart_proxy_dynflow_core.rb', 'config/settings.yml.example',
+                          'bundler.d/*', 'lib/smart_proxy_dynflow_core/*', 'LICENSE', 'Gemfile',
```
ares: this way I suppose you're missing `lib/smart_proxy_dynflow_core/.rb` (which I see is listed above), why not using the same approach as in smart_proxy_remote_execution_ssh with /**/*? anyway that's minor thing

Gemfile
adamruzicka: Beat me to it :)

smart_proxy_dynflow_core.gemspec
adamruzicka: I figured out it was the space in `{bundler.d, lib/smart_proxy_dynflow_core}'` which broke it

config/settings.yml.example
adamruzicka: `Rack::Server` accepts `Host` and `Port` as options for the webserver and so I reused those names in the settings

lib/smart_proxy_dynflow_core/settings.rb
```diff
@@ -0,0 +1,59 @@
...
+      super(DEFAULT_SETTINGS.merge(settings))
+    end
+
+    def self.instance
+      SmartProxyDynflowCore::SETTINGS
+    end
+
+    def self.load_global_settings(path)
+      if File.exists? File.join(path)
+        YAML.load(File.read(path)).each do |key, value|
```
ares: `YAML.load_file(path)` is a shortcut (on more places in the PR)

ares:
General notes from testing so far:

**logging**
no core is not running and smart proxy does, when I run a command I get a failure and in job output I see "Failed to initialize command: Dynflow::ExecutionPlan::Steps::Error - The only applicable proxy rexproxy.example.tst is down" which is not true. Also in proxy log even with DEBUG log level there's just this

```
D, [2016-05-12T11:49:15.025330 #30395] DEBUG -- : accept: 192.168.122.1:35214
D, [2016-05-12T11:49:15.027093 #30395] DEBUG -- : Rack::Handler::WEBrick is invoked.
D, [2016-05-12T11:49:15.029951 #30395] DEBUG -- : verifying remote client 192.168.122.1 against trusted_hosts rexproxy.example.tstforeman.example.tst
D, [2016-05-12T11:49:15.032347 #30395] DEBUG -- : require_ssl_client_verification: skipping, non-HTTPS request
D, [2016-05-12T11:49:15.216034 #30395] DEBUG -- : close: 192.168.122.1:35214
```

I'd like to see that some proxying request was received and it's status in debug log, plus in case of connection failure we should log some error.

**gemspec and bundler_ext**
I created Gemfile.in in lib/ where it's expected but then I'm getting following error (gemspec is missing in that directory)
```
[!] There was an error parsing `Gemfile.in`: There are no gemspecs at /opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/lib. Bundler cannot continue.
 #  from /opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/lib/Gemfile.in:3
 #  -------------------------------------------
 #  
 >  gemspec :name => 'smart_proxy_dynflow_core'
 #  
 #  -------------------------------------------
        from /opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/lib/Gemfile.in:3:in `eval_gemfile'
```

I'm not sure how this normally works but I suppose we'll have to use bundler_ext and generate Gemfile.in during packaging.

**configuration**
how do I specify which config file should be loaded, I expect we don't want to configure it at `/opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/config/settings.yml`

with setting.yml missing I'm getting ugly error, would it be possible to add some check?
```
/opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/bin/smart_proxy_dynflow_core.rb
"/opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/lib/Gemfile.in"
/opt/rh/ruby193/root/usr/local/share/gems/gems/sequel-4.34.0/lib/sequel/adapters/sqlite.rb:109:in `initialize': SQLite3::CantOpenException: unable to open database file (Sequel::DatabaseConnectionError)
```
**config load error**
getting following error when starting the core, it seems that SETTINGS does not implement it

```
/opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/bin/smart_proxy_dynflow_core.rb
/opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/lib/smart_proxy_dynflow_core/settings.rb:32:in `block in load_global_settings': undefined method `[]=' for #<SmartProxyDynflowCore::Settings database="/var/lib/foreman-proxy/dynflow/dynflow.sqlite", callback_url="https://127.0.0.1:8443", console_auth=true, foreman_url="http://127.0.0.1:3000", Host="127.0.0.1", Port="8008", use_https=false, ssl_ca_file=nil, ssl_private_key=nil, ssl_certificate=nil, standalone=false, plugins={}> (NoMethodError)
        from /opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/lib/smart_proxy_dynflow_core/settings.rb:31:in `each'
        from /opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/lib/smart_proxy_dynflow_core/settings.rb:31:in `load_global_settings'
        from /opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/lib/smart_proxy_dynflow_core/launcher.rb:20:in `load_settings!'
        from /opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/lib/smart_proxy_dynflow_core/launcher.rb:12:in `start'
        from /opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/lib/smart_proxy_dynflow_core/launcher.rb:8:in `launch!'
        from /opt/rh/ruby193/root/usr/local/share/gems/gems/smart_proxy_dynflow_core-0.0.7/bin/smart_proxy_dynflow_core.rb:7:in `<main>'
```

adamruzicka:
> logging

agreed, needs more logging. However the "Failed to initialize command: Dynflow::ExecutionPlan::Steps::Error - The only applicable proxy rexproxy.example.tst is down" part will have to be handled in REX. 

> gemspec and bundler_ext

will have to read up more on this

> how do I specify which config file should be loaded?

There's no way to configure a config file now, would a parameter for it be enough? Maybe a list of common places where it could be too?

> with setting.yml missing I'm getting ugly error, would it be possible to add some check?
Will do

> config load error
Apparently `OpenStruct` doesn't support hash-like access on 1.9.3, I'll have to come up with a workaround

ares:
>> how do I specify which config file should be loaded?

>There's no way to configure a config file now, would a parameter for it be enough? Maybe a list of common places where it could be too?

if you implement the list then '/etc/smart_proxy_dynflow_core/' sounds as a good candidate. The argument might work as well as long as we put it into a service file.

ares:
btw this seems to work for me as a simple and ugly workaround for OpenStruct issue...
```ruby
    def [](key)
      self.send key
    end

    def []=(key, value)
      self.send "#{key}=", value
    end
```

Gemfile
ares: ok, currently we pack this Gemfile in the gem which is causing problems, maybe we should wrap this in some $var condition and skip it in core gemspec

config/settings.yml.example
ares: so this file is not config for core, I suppose we need one for smart_proxy_dynflow with
`:core_url: http://localhost:8008` and standard options like enabled

lib/smart_proxy_dynflow/callback.rb
```diff
@@ -4,37 +4,37 @@ module Proxy
...
-          plan_self(:callback => callback, :data => data)
+      class Core < Proxy::HttpRequest::ForemanRequest
+        def uri
+          @uri ||= URI.parse Proxy::Dynflow::Plugin.settings.core_url
         end
 
-        def run
-          Callback::Request.send_to_foreman_tasks(input[:callback], input[:data])
+        def relay(request, from, to)
+          path = request.env['REQUEST_PATH'].gsub(from, to)
```
ares: `request.env['REQUEST_PATH']` returns '/' on my setup, `request.path` seems to work

config/settings.yml.example
adamruzicka: This file is for core, the one for `smart_proxy_dynflow` (the proxy plugin itself) is in `settings.d/dynflow.yml.example`. This should be explicitly mentioned somewhere

lib/smart_proxy_dynflow_core/settings.rb
```diff
@@ -0,0 +1,59 @@
...
+        SETTINGS[key] = plugin.settings[key]
+      end
+      SETTINGS.plugins.values.each { |plugin| plugin.load_settings_from_proxy }
+    end
+
+    def self.load_plugin_settings(path)
+      settings = YAML.load(File.read(path))
+      name = File.basename(path).gsub(/\.yml$/, '')
+      if SETTINGS.plugins.key? name
+        settings = SETTINGS.plugins[name].to_h.merge(settings)
```
ares: in Ruby 1.9.3, #to_h returns `nil` (as well as marshal_dump) :-(

adamruzicka:
> config location

currently it searches for `settings.yml` in `/etc/smart_proxy_dynflow_core`, `~/.config/smart_proxy_dynflow_core`, `config` in source code directory and optionally in the directory provided as argument to the `--config-dir`. It loads them all in this order with the later overriding the earlier. If the `-1` flag is supplied, it processes the locations in reverse order, stopping after a successful load

> config load error and friends on 1.9.3

should be fixed

> proper logging and bundler_ext/in

to be done

bin/smart_proxy_dynflow_core
```diff
@@ -3,5 +3,22 @@ require 'dynflow'
 require 'rack'
 require 'smart_proxy_dynflow_core/launcher'
 require 'yaml'
+require 'optparse'
```
ares: it seems we have clamp packaged in SCL and nonSCL form, maybe we could use that instead of optparse?

bin/smart_proxy_dynflow_core
```diff
@@ -3,5 +3,22 @@ require 'dynflow'
...
 
-SmartProxyDynflowCore::Launcher.launch!
+options = {}
+OptionParser.new do |opts|
+  opts.on('-c', '--config-dir CONFIG_DIR', String, 'Directory to load settings from') do |value|
+    options[:config_dir] = value
+  end
+
+  opts.on('-1', '--one-config', String, 'Do not load more than 1 config') do |value|
```
ares: in what scenario is this useful? it looks a bit confusing that I can combine multiple config files

bin/smart_proxy_dynflow_core
adamruzicka: `optparse` is in ruby's stdlib so we won't need to package it. `clamp` felt like an overkill for this use case.

bin/smart_proxy_dynflow_core
adamruzicka: Since it by default traverses all the possible paths where the config could be and merges all the discovered configs I thought it would be nice to have a way to say "load the config from $place and nothing else"

stbenjam:
It's not clear to me how the app is supposed to be started stand-alone:

```
$ bundle exec bin/smart_proxy_dynflow_core
/home/stbenjam/.rvm/gems/ruby-2.0.0-p643/bundler/gems/smart-proxy-185cfbb776aa/lib/proxy/plugin.rb:63:in `plugin': uninitialized constant Proxy::Plugins (NameError)
```

What am I missing?

adamruzicka:
@stbenjam You either have to remove the `gem 'smart_proxy_dynflow', :path => '.'` from the Gemfile or add the `smart-proxy` gem to the bundle (just a guess, can't test it right now, but I vaguely remember doing it in my Gemfile.local.rb). I'm not sure how to handle the shared Gemfile properly

stbenjam:
Hm, why is smart_proxy_dynflow in the Gemfile at all?

adamruzicka:
Because both the `_core` and regular version share it and both gems need to be in the bundle for example when running tests

stbenjam:
You could restrict the smart_proxy_dynflow requirement to the test group, for example.  

It's odd the gemfile says for example `gemspec :name => 'smart_proxy_dynflow_core'`, but you're using this same gemfile for smart_proxy_dynflow.

stbenjam:
I opened a PR against your branch to add the service files: https://github.com/adamruzicka/smart_proxy_dynflow/pull/1

adamruzicka:
@stbenjam Thanks! Left one inline comment there

iNecas:
I've squashed the commits, pushed to master and release 0.1.0 version for both proxy and core gems

iNecas:
Thanks everyone for this, let's do additional improvements in another PRs

PR #17, opened 2016-06-07
fixes #15335 - fixes required for packaging

stbenjam:
This looks for the Gemfile in the root directory, adds the gemspec to
the gem, and loads plugins like smart_proxy_remote_execution_ssh_core
from the bundler.d/* directory.


stbenjam:
ssh core part: https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/24

adamruzicka:
[This](https://github.com/stbenjam/smart_proxy_dynflow/blob/f644e277c63d560abc603d8926bed343d6c0b0b6/bundler.d/dynflow.rb#L1) should probably get removed, since we don't want to try to load the non-core part in core

stbenjam:
We still need it, it's used to put it in the foreman-proxy bundler.d for non-core part.  I can update the gemspec to not include it in core, but I guess it will still affect running from source.

Any ideas?

stbenjam:
I guess maybe we could call the directory something else

adamruzicka:
The same problem is in `smart_proxy_remote_execution_ssh`. Having separate `bundler.d`s seems like the cleanest solution

ares:
sounds reasonable, `bundler.d.plugins` maybe?

stbenjam:
Ok, how about this? I did bundler.plugins.d, looks better with the .d on the end to me.  I also moved them in https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/24, even though it doesn't need to be, I think it makes sense if we use the same directory structure for REX stuff.

adamruzicka:
I like that :+1:

adamruzicka:
So just to be sure.
```shell
ssh=smart_proxy_remote_execution_ssh
dynflow=smart_proxy_dynflow
```
file | will go to
-----|-----------
`$dynflow/bundler.plugins.d/dynflow.rb` | `smart-proxy/bundler.d/`
`$ssh/bundler.plugins.d/remote_execution_ssh.rb` | `smart-proxy/bundler.d/`
`$ssh/bundler.plugins.d/remote_execution_ssh_core.rb` |`$dynflow/bundler.d/`

The `$dynflow/bundler.d` directory will be created during the package installation?

stbenjam:
@adamruzicka Yup, exactly that, I'll handle moving them around in the packaging so they go to the right places.

adamruzicka:
@stbenjam great, ACK

Do we have to wait for green tests with this? The proposed changes should have no impact on test results

stbenjam:
No, I don't think so, I'll merge and release 0.1.1 so I can get new packages built

PR #20, opened 2016-06-16
Fixes #15424 - Be able to daemonize

iNecas:
https://github.com/theforeman/smart_proxy_dynflow/pull/19 + a fix for initialization
after daemonizing.

PR #21, opened 2016-06-16
Fixes #15442 - use separate settings for client certs

iNecas:
They might differ form the certs used for https.

Also, we now detect the need for certs based on the schema of the callback url.

I've updated the settings.yml.example to start with http by default, for
development purposes, not expecting the devels to have the certs in place.

The certs part in setting file I used when testing with Katello
looked something like this:

```
:ssl_ca_file:  /etc/foreman-proxy/ssl_ca.pem
:ssl_private_key: /etc/foreman-proxy/ssl_key.pem
:ssl_certificate: /etc/foreman-proxy/ssl_cert.pem

:foreman_ssl_ca:  /etc/foreman-proxy/foreman_ssl_ca.pem
:foreman_ssl_key: /etc/foreman-proxy/foreman_ssl_key.pem
:foreman_ssl_cert: /etc/foreman-proxy/foreman_ssl_cert.pem
```

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -2,8 +2,15 @@
 # Path to dynflow database, leave blank for in-memory non-persistent database
 :database:
 
-# URL of the smart proxy, used for reporting back
-:callback_url: 'https://localhost:8443'
+# URL of the foreman, used for reporting back
+:foreman_url: 'http://localhost:3000'
+
+# SSL settings for client authentication against Foreman
+# :foreman_ssl_ca: ssl/foreman_ca.pem
+# :foreman_ssl_key: ssl/foreman_key.pem
+# :foreman_ssl_cert: ssl/foreman_cert.pem
+
+:console_auth: false
 
 # Set to true to make the core fork to background after start
 # :daemonize: false
@@ -15,7 +22,7 @@
 # Listen on port
 :port: 8008
 
-# SSL settings for client authentication against smart proxy.
+# SSL settings for running core as https service
 # :use_https: false
 # :ssl_ca_file: ssl/ca.pem
 # :ssl_private_key: ssl/localhost.pem
@@ -26,4 +33,3 @@
 
 # Log level, one of UNKNOWN, FATAL, ERROR, WARN, INFO, DEBUG
 # :log_level: ERROR
-
--- a/lib/smart_proxy_dynflow_core/callback.rb
+++ b/lib/smart_proxy_dynflow_core/callback.rb
@@ -10,15 +10,15 @@ module SmartProxyDynflowCore
   module Callback
     class Request
       def callback(payload)
-        response = callback_resource.post payload
-        if response.code != 200
-          raise "Failed performing callback to smart proxy: #{response.code} #{response.body}"
+        response = callback_resource.post(payload, :content_type => :json)
+        if response.code.to_s != "200"
+          raise "Failed performing callback to Foreman server: #{response.code} #{response.body}"
         end
         response
       end
 
-      def self.callback(callback, data)
-        self.new.callback(self.prepare_payload(callback, data))
+      def self.send_to_foreman_tasks(callback_info, data)
+        self.new.callback(self.prepare_payload(callback_info, data))
       end
 
       private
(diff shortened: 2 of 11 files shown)


iNecas:
Also, one needs to make sure the urls for callbacks use fqdn, not localhost, in order the verification to work.

config/settings.yml.example
```diff
@@ -3,7 +3,12 @@
 :database:
 
 # URL of the smart proxy, used for reporting back
-:callback_url: 'https://localhost:8443'
+:callback_url: 'http://localhost:9292'
+
+# SSL settings for client authentication against smart proxy.
+# :ssl_ca_file: ssl/foreman_ca.pem
+# :ssl_private_key: ssl/foreman_key.pem
+# :ssl_certificate: ssl/foreman_cert.pem
```
stbenjam: I'm a little confused, these are the same settings names as below in the file

stbenjam: These should beforeman_ssl_private_key, foreman_ssl_certificate and foreman_ssl_ca_file, right?

config/settings.yml.example
```diff
@@ -3,7 +3,12 @@
 :database:
 
 # URL of the smart proxy, used for reporting back
-:callback_url: 'https://localhost:8443'
+:callback_url: 'http://localhost:9292'
```
adamruzicka: Why `9292`? Isn't proxy listening on `8443` for https and `8000` for http by default?

lib/smart_proxy_dynflow_core/callback.rb
```diff
@@ -33,13 +33,13 @@ def callback_resource
       end
 
       def ssl_options
-        return {} unless Settings.instance.use_https
-        client_key = File.read  Settings.instance.ssl_private_key
-        client_cert = File.read Settings.instance.ssl_certificate
+        return {} unless URI.parse(Settings.instance.callback_url).scheme == 'https'
```
adamruzicka: If we're going to decide whether to use http or https based on parsing the uri, we could (and should) drop the `use_https` from config and settings

stbenjam:
Needs rebase as well after merging #20

lib/smart_proxy_dynflow_core/callback.rb
iNecas: I don't think so, we need the use_https do decide, if we the core itself as http or https: it's not related directly on what to use for authenticating against callback

config/settings.yml.example
iNecas: Right: my bed. updated

config/settings.yml.example
iNecas: It was my default for running proxy locally. Anyways, thinking more about the ssl authentication and how the callback works, I've at the end changed the behavior in a way, where the core doesn't send the callback through the proxy, but rather calls foreman directly: there is no need to go though proxy for this part + if we did, we would need to add the  proxy to it's trusted hosts so that the call is accepted.

iNecas:
Rebased and tested with katello on rhel6, works fine

stbenjam:
use_https needs removing from the sample config too, otherwise looks ok to me.  I'll test it today

iNecas:
Use_https is still used: the change is before, it was used also when determining how to do callbacks back to Foreman after the task finishes. Right now, it's used only to determine if the core runs as http or https.

test/api_test/test_helper.rb
```diff
@@ -9,5 +9,7 @@
 
 require 'smart_proxy_dynflow'
 
+Proxy::Dynflow::Plugin.load_test_settings({})
```
iNecas: Our test suite was not gree due to this line missing, so the default config was not loaded with recent proxy

iNecas:
Tests should get green after recent change in test_helper, I've sent a build manually http://ci.theforeman.org/job/test_proxy_plugin_pull_request/116/

stbenjam:
Thanks @iNecas, ACK from me.  Merge away

iNecas:
Few updates added that came out of testing:

* I've moved the console_auth setting to the core: it seems it was missed at the original PR
* limit the accepted certificate in core client authentication only to one of the proxy - otherwise, any certificate issued by the truste CA could be used for the authentication, which is not what we want to have. With this change, effectively only the proxy process can access it

iNecas:
Tests running here http://ci.theforeman.org/job/test_proxy_plugin_pull_request/117/

iNecas:
Tests are green. @adamruzicka would you mind giving it your blessing before merge?

adamruzicka:
Do we need to keep separate `callback_url` and `foreman_url` in settings? In standalone mode the core called back to foreman through the proxy (`callback_url` here) or directly to foreman when running as part of the proxy (`foreman_url`). Now with this patch it will always call back directly to foreman so we could probably drop one of those.

Otherwise looks good to me

iNecas:
@adamruzicka good idea, done. New tests running here http://ci.theforeman.org/job/test_proxy_plugin_pull_request/119/

config/settings.yml.example
```diff
@@ -2,8 +2,15 @@
...
 :database:
 
-# URL of the smart proxy, used for reporting back
-:callback_url: 'https://localhost:8443'
+# URL of the foreman, used for reporting back
+:foreman_url: 'http://localhost:3000'
+
+:console_auth: false
+
+# SSL settings for client authentication against smart proxy.
```
stbenjam: These certs are used to connect back to the foreman then? This comment should be changed then

iNecas: Right, I'will change that to `'SSL settings for client authentication against Foreman.'` and merge the PR

iNecas: Right, updated

stbenjam:
ACK

adamruzicka:
ACK from me too

PR #26, opened 2016-09-02
Refs #16058 - integrate foreman-tasks-core

iNecas:
This allows any gem using foreman-tasks-core to use both foreman and
smart proxy as the process for running it

### DIFF
--- a/lib/smart_proxy_dynflow/plugin.rb
+++ b/lib/smart_proxy_dynflow/plugin.rb
@@ -17,5 +17,14 @@ class Proxy::Dynflow
     requires :foreman_proxy, ">= 1.12.0"
     default_settings :core_url => 'http://localhost:8008'
     plugin :dynflow, Proxy::Dynflow::VERSION
+
+    after_activation do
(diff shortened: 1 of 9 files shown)


lib/smart_proxy_dynflow/plugin.rb
```diff
@@ -17,5 +17,15 @@ class Plugin < Proxy::Plugin
...
     default_settings :core_url => 'http://localhost:8008'
     plugin :dynflow, Proxy::Dynflow::VERSION
+
+    after_activation do
+      begin
+        require 'smart_proxy_dynflow_core'
+      rescue LoadError => e
+        raise e
+        # Dynflow core is not available in the proxy, will be handled
+        # by standalone Dynflow core
```
stbenjam: Should this be reraising? The comment makes me think otherwise

iNecas: Right, forgot this after debugging

iNecas:
[test]

PR #27, opened 2016-10-25
Fixes #17078 - Allow disabling certain ciphers

adamruzicka:


iNecas:
Works as expected, thanks @adamruzicka

PR #28, opened 2016-11-07
Fixes #17249 - Support regular authentication for allin1 setup

ares:


iNecas:
[test]

ares:
[test]

iNecas:
[test] again, there was a syntax error in the jenkins job defiinition causing the builds to fail

iNecas:
Tests are failing

adamruzicka:
@ares What's the status here?

ares:
I'd like to look at this sooner or later, I think we should production-like setups even if we're running dev env on single host.

ares:
sorry it took me a while, this now works with both separate core and all in one setup, @iNecas @adamruzicka please take a look before we forget what id does

ares:
Requires https://github.com/theforeman/smart-proxy/pull/550, both setups tested and it seems to work fine.

lib/smart_proxy_dynflow_core/authorization_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+        if defined?(::ForemanTasksCore)
+          auth = request.env['HTTP_AUTHORIZATION']
+          basic_prefix = /\ABasic /
+          if !auth.to_s.empty? && auth =~ basic_prefix &&
+              ForemanTasksCore::OtpManager.authenticate(auth.gsub(basic_prefix, ''))
+            Log.instance.debug('authorized with token')
+            @authenticated = true
+          end
+        end
+        halt 403, MultiJson.dump(:error => 'Invalid username or password supplied')
```
adamruzicka: Still not quite there yet. Even if the token based auth succeeds, `@authenticated` is set to true on line 34, but the processing still hits this `halt`, resulting in `403 Forbidden`

ares: right, adding return after @authenticated assignment should do the trick

ares:
> sorry it took me a while, this now works with both separate core and all in one setup, @iNecas @adamruzicka please take a look before we forget what id does

I already did, while I think it's still valuable, I'm closing this for inactivity

PR #30, opened 2017-04-11
Fixes #19224 - Reopen logs on SIGUSR1

adamruzicka:


lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -40,6 +40,20 @@ def log_file
...
+      super(*args)
+    end
+
+    def roll_log
+      @roll_log = true
+    end
+
+    def handle_log_rolling
+      @roll_log = false
+      self.reopen
```
iNecas: reopen is not defined in ruby <= 2.3.0

lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -40,6 +40,25 @@ def log_file
...
+      super(*args)
+    end
+
+    def roll_log
+      @roll_log = true
+    end
+
+    def handle_log_rolling
+      @roll_log = false
+      @fd.reopen unless log_file.kind_of? IO
```
iNecas: Where is log_file defined? Ans why the `unless` condition?

adamruzicka: [log.rb#L34](https://github.com/adamruzicka/smart_proxy_dynflow/blob/99a8eb97c9e054c07577d9df79ea6ff9ba56ffad/lib/smart_proxy_dynflow_core/log.rb#L34)

There are two possibilities either it is a string which we treat as a filesystem path or it is stdout. Reopening stdout on logrotate doesn't make much sense

iNecas: This is defined on class, while we are on instance: this doesn't work. Why not just saving the @file in initializer in case it's string and using it as condition + reopen needs file and mode specified, otherwise it doesn't work.

lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -40,6 +40,11 @@ def log_file
       end
     end
 
+    def initialize(file, *rest)
+      @fd = file.kind_of?(IO) ? file : File.open(file, 'a')
```
iNecas: The `@fd` should be set with `sync = true`, to write the the lines to the file imediatelly

lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -8,7 +8,7 @@ class Log < ::Logger
     class << self
       def instance
         if @logger.nil?
-          @logger = Logger.new log_file
+          @logger = self.new log_file
```
iNecas: Seems the log_file is being written to by webrick access log (see the `launcher.rb`) which breaks the whole thing. We need to pass something else to the webrick access log (it seems it accepts the logger) instance.

iNecas:
Works fine now, thanks @adamruzicka

PR #31, opened 2017-04-12
Fixes #18449 - Limit number of opened files to 65536 for EL6

adamruzicka:
the default is set to 1024


iNecas:
Thanks @adamruzicka

PR #34, opened 2017-05-25
Fixes #19639 - Fix loading settings from smart-proxy

adamruzicka:
this only affects the setup where dynflow executor runs inside smart-proxy


mmoll:
fixes the problem for me 👍

adamruzicka:
Wow, that was quick, thanks @mmoll

iNecas:
Thanks @adamruzicka

PR #36, opened 2017-08-17
Fixes #20624 - Register dead letter handler from foreman tasks core

adamruzicka:
requires
- [x] https://github.com/theforeman/foreman-tasks/pull/276
- [x] https://github.com/Dynflow/dynflow/pull/245

### DIFF
--- a/lib/smart_proxy_dynflow_core/core.rb
+++ b/lib/smart_proxy_dynflow_core/core.rb
@@ -43,6 +43,9 @@ module SmartProxyDynflowCore
         config.logger_adapter = logger_adapter
         config.persistence_adapter = persistence_adapter
         config.execution_plan_cleaner = execution_plan_cleaner
+        # TODO: There has to be a better way
+        matchers = config.silent_dead_letter_matchers.call().concat(self.class.plugin_matchers)
+        config.silent_dead_letter_matchers = matchers
         yield config if block_given?
       end
     end
@@ -71,6 +74,14 @@ module SmartProxyDynflowCore
         @instance = Core.new
         after_initialize_blocks.each { |block| block.call(@instance) }
         @instance
+      end
+
+      def plugin_matchers
+        @matchers ||= []
+      end
+
+      def register_plugin_matchers(matchers)
+        plugin_matchers.concat matchers
       end
 
       def web_console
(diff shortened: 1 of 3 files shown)


iNecas:
Not sure why jenkins is not picking up the task

iNecas:
[test]

iNecas:
Tests are failing, as we renamed the `DeadLetterHandler` to `DeadLetterSilencer`

adamruzicka:
Fixed in https://github.com/theforeman/foreman-tasks/pull/281

smart_proxy_dynflow_core.gemspec
```diff
@@ -30,8 +30,8 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency('rack-test', '~> 0')
   gem.add_development_dependency('rubocop', '0.32.1')
 
-  gem.add_runtime_dependency('dynflow', "~> 0.8.27")
-  gem.add_runtime_dependency('foreman-tasks-core', '>= 0.1.5')
+  gem.add_runtime_dependency('dynflow', "~> 0.8.29")
+  gem.add_runtime_dependency('foreman-tasks-core', '>= 0.1.6')
```
iNecas: Please update to 0.1.7, as the fix for silencer is included there.

lib/smart_proxy_dynflow_core/core.rb
```diff
@@ -73,6 +76,14 @@ def ensure_initialized
         @instance
       end
 
+      def plugin_matchers
```
iNecas: From the name it's not clear, what kind of matchers it is. What about `silencer_matchers`?

lib/smart_proxy_dynflow_core/core.rb
```diff
@@ -73,6 +76,14 @@ def ensure_initialized
...
+        @matchers ||= []
+      end
+
+      def register_plugin_matchers(matchers)
```
iNecas: Similarly to `plugin_matchers`, thinking about improving the naming. What about `register_silencer_matchers`, as we don't really care if it's from plugins, but it should be clean it's about silencing

iNecas:
Tested and works well. Thanks @adamruzicka

PR #37, opened 2017-08-17
Fixes #20626 - Periodic execution plan cleaner

adamruzicka:


iNecas:
One thing that came out of the testing: the cleaner should emit some info log messages about it's activity: it can be very useful

iNecas:
The log needs to probably go into Dynflow

lib/smart_proxy_dynflow_core/settings.rb
```diff
@@ -40,6 +40,7 @@ class Settings < OpenStruct
         :plugins => {},
         :pid_file => '/var/run/foreman-proxy/smart_proxy_dynflow_core.pid',
         :daemonize => false,
+        :execution_plan_cleaner_age => 60 * 60 * 24,
```
iNecas: You need to add this also to `PLUGIN_SETTINGS` for the code to properly load the data from settings

iNecas:
Tested and works well, after doing some changes mentioned in the review

adamruzicka:
Addressed comments, logging is handled in https://github.com/Dynflow/dynflow/pull/248

iNecas:
I don't see the code that would address the comments: forgot to push?

adamruzicka:
Pushed now...

settings.d/dynflow.yml.example
```diff
@@ -2,3 +2,4 @@
 :enabled: true
 :database: /var/lib/foreman-proxy/dynflow/dynflow.sqlite
 :core_url: 'http://127.0.0.1:8008'
+:cleaner_max_age: 86400 # 1 day
```
iNecas: Why this one?

iNecas:
I've released dynflow 0.8.27, please bump the dependency here.

iNecas:
Test failures unrelated. Merging

iNecas:
Tested and works great. Thanks @adamruzicka

PR #40, opened 2017-11-15
Provide API for querying task states

adamruzicka:


iNecas:
Thanks @adamruzicka, could you release a new version to not forget about this.

PR #41, opened 2018-01-03
pin gem versions for older Rubies

mmoll:
also removes pinning for Ruby <2.0

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,12 +9,26 @@ end
 group :test do
   gem 'smart_proxy_dynflow', :path => '.'
   gem 'smart_proxy', :git => "https://github.com/theforeman/smart-proxy", :branch => "develop"
+
+  if RUBY_VERSION < '2.1'
+    gem 'public_suffix', '< 3'
+  else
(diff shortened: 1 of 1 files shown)


mmoll:
http://ci.theforeman.org/job/test_proxy_plugin_pull_request/250/ 💚

Gemfile
```diff
@@ -9,12 +9,26 @@ end
 group :test do
   gem 'smart_proxy_dynflow', :path => '.'
   gem 'smart_proxy', :git => "https://github.com/theforeman/smart-proxy", :branch => "develop"
+
+  if RUBY_VERSION < '2.1'
+    gem 'public_suffix', '< 3'
+  else
+    gem 'public_suffix'
```
adamruzicka: Why do we suddenly need this?

mmoll: webmock -> addressable -> public_suffix

adamruzicka: Oh, I see. Thanks for clarifying it

adamruzicka:
Thanks @mmoll !

PR #42, opened 2018-01-04
Fixes #22141 - Remove sinatra/rack dependencies

dLobatog:
Without this, I'm getting the following error when running `bundle update` on smart-proxy (develop branch). I suppose the plugin can simply rely on sinatra/rack being provided by smart-proxy.

```
Resolving dependencies......
Bundler could not find compatible versions for gem "rack-protection":
  In Gemfile:
    smart_proxy was resolved to 1.17.0.develop, which depends on
      sinatra was resolved to 1.4.6, which depends on
        rack-protection (= 2.0.0)

    smart_proxy_dynflow_core was resolved to 0.1.8, which depends on
      sinatra (~> 1.4) was resolved to 1.4.6, which depends on
        rack-protection (~> 1.4)
Bundler could not find compatible versions for gem "tilt":
  In Gemfile:
    smart_proxy was resolved to 1.17.0.develop, which depends on
      sinatra was resolved to 1.4.0, which depends on
        tilt (~> 2.0)

    smart_proxy_dynflow_core was resolved to 0.1.8, which depends on
      sinatra (~> 1.4) was resolved to 1.4.0, which depends on
        tilt (>= 1.3.4, ~> 1.3)
```

### DIFF
--- a/smart_proxy_dynflow_core.gemspec
+++ b/smart_proxy_dynflow_core.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency('foreman-tasks-core', '>= 0.1.7')
   gem.add_runtime_dependency('sequel')
   gem.add_runtime_dependency('sqlite3')
-  gem.add_runtime_dependency('sinatra', '~> 1.4')
+  gem.add_runtime_dependency('sinatra')
   gem.add_runtime_dependency('rack')
   gem.add_runtime_dependency('rest-client')
 end


iNecas:
We can't, as this is for smart_proxy_dynflow_core, which can run separately but needs sinatta for dynflow console and internal api

dLobatog:
@iNecas How do you run it then with smart-proxy nightly?

adamruzicka:
I was able to reproduce the issue @dLobatog describes when running dynflow core as part of smart proxy.

iNecas:
@dLobatog ping

smart_proxy_dynflow_core.gemspec
```diff
@@ -34,7 +34,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency('foreman-tasks-core', '>= 0.1.7')
   gem.add_runtime_dependency('sequel')
   gem.add_runtime_dependency('sqlite3')
-  gem.add_runtime_dependency('sinatra', '~> 1.4')
+  gem.add_runtime_dependency('sinatra')
```
ekohl: Should this be ```>= 1.4``` instead?

iNecas: I think we should be ok with whatever we get from the proxy, so I'm ok with this way

iNecas:
Thanks @dLobatog

smart_proxy_dynflow_core.gemspec
ekohl: The core part doesn't have to run with the proxy I think but I agree that chance is remote.

PR #45, opened 2018-02-19
Refs ##22513 - Remove the file limit

chris1984:


adamruzicka:
If you take a look at the issue[1] for which the limit was added you will find out it was actually done to increase the limit from system's defaults, which, at least on EL7, is much lower than 65k. Reverting it probably won't help with supporting large client installs.

[1] - http://projects.theforeman.org/issues/18449

chris1984:
@adamruzicka we are adding it here https://github.com/theforeman/puppet-foreman_proxy/pull/409

sean797:
@chris1984 what about EL6? Are we still supporting EL6 Smart Proxies?

https://github.com/theforeman/smart_proxy_dynflow/blob/432f2908d1d7b1b6eb46ec0ad4a3812575c72c9b/deploy/smart_proxy_dynflow_core.init#L18

chris1984:
@sean797 i thought support for el6 was long gone? If so I can remove it out of there?

adamruzicka:
Oh, right, sorry, I missed that PR. After reading though it, this change makes sense.

iNecas:
So I guess this is waiting until theforeman/puppet-foreman_proxy#409 gets merged?

chris1984:
@iNecas can you merge the other is now in, thanks

iNecas:
Sure, thanks @chris1984

PR #48, opened 2018-03-19
Fixes #22935 - Initialize Dynflow just before the web server starts

adamruzicka:


iNecas:
Nice finding. Was trying to reproduce the original issue, but I believe this will help with the issue.

PR #50, opened 2018-05-10
add EnvironmentFile in systemd unit

hartmel:
Smart proxy has already this systemd EnvironmentFile directive and can be use as workaround for SSH remote execution.
Because of shell defined to /bin/false for foreman-proxy user, SSH command (ansible, remote exec) may fail without trying to connect. The environment file allow to define an environment variable like SHELL=/bin/bash


stbenjam:
[test]

deploy/smart_proxy_dynflow_core.service
```diff
@@ -8,6 +8,7 @@ Type=forking
 User=foreman-proxy
 PIDFile=/var/run/foreman-proxy/smart_proxy_dynflow_core.pid
 ExecStart=/usr/bin/smart_proxy_dynflow_core -d -p /var/run/foreman-proxy/smart_proxy_dynflow_core.pid
+EnvironmentFile=-/etc/sysconfig/smart_proxy_dynflow_core
```
akofink: Is the `-` supposed to be there before the path?

stbenjam: Yea it's a systemd thing, it means ignore errors in the file isn't there I think

hartmel: exactly

iNecas:
Thanks @hartmel

PR #55, opened 2018-10-09
Fixes #25403 - API for batch triggering

adamruzicka:


lib/smart_proxy_dynflow_core/api.rb
```diff
@@ -21,6 +21,14 @@ class Api < ::Sinatra::Base
       MultiJson.dump(result)
     end
 
+    post "/tasks/batch/?" do
+      params = MultiJson.load(request.body.read)
+      triggered = world.trigger(::Dynflow::Utils.constantize(params['action_name']),
+                                callback_host(params, request),
+                                params['action_input'])
+      world.persistence.load_execution_plan(triggered.id).actions.first.input[:result].to_json
```
iNecas: I wonder if it's safe to expect that there always will be at least one action, and that this action will have `:result` in the input. Couldn't we rely on some simpler response?

PR #56, opened 2018-10-23
Fixes #25239 - improved reporting exceptions

lzap:
The issue is fixed on smart proxy side by

https://github.com/theforeman/smart-proxy/pull/619

but this patch improves behavior.

### DIFF
--- a/lib/smart_proxy_dynflow_core/log.rb
+++ b/lib/smart_proxy_dynflow_core/log.rb
@@ -54,13 +54,33 @@ module SmartProxyDynflowCore
       end
     end
 
+    class ProxyStructuredFormater < ::Dynflow::LoggerAdapters::Formatters::Abstract
+      def call(_severity, _datetime, _prog_name, message)
+        if ::Exception === message
+          subject = "#{message.message} (#{message.class})"
+          if @base.respond_to?(:exception)
+            @base.exception("Error details", message)
+            subject
+          else
+            "#{subject}\n#{message.backtrace.join("\n")}"
+          end
+        else
+          message
+        end
+      end
+
+      def format(message)
+        call(nil, nil, nil, message)
+      end
+    end
+
     class ProxyAdapter < ::Dynflow::LoggerAdapters::Simple
-      def initialize(logger, level = Logger::DEBUG, formatters = [::Dynflow::LoggerAdapters::Formatters::Exception])
+      def initialize(logger, level = Logger::DEBUG, formatters = [])
         @logger           = logger
         @logger.level     = level
-        @logger.formatter = method(:formatter).to_proc
-        @action_logger    = apply_formatters ProgNameWrapper.new(@logger, ' action'), formatters
-        @dynflow_logger   = apply_formatters ProgNameWrapper.new(@logger, 'dynflow'), formatters
+        @logger.formatter = ProxyStructuredFormater.new(@logger)
+        @action_logger    = apply_formatters(ProgNameWrapper.new(@logger, ' action'), [ProxyStructuredFormater])
+        @dynflow_logger   = apply_formatters(ProgNameWrapper.new(@logger, 'dynflow'), [ProxyStructuredFormater])
       end
     end
   end
--- a/smart_proxy_dynflow_core.gemspec
+++ b/smart_proxy_dynflow_core.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency('rack-test', '~> 0')
   gem.add_development_dependency('rubocop', '~> 0.52.1')
 
-  gem.add_runtime_dependency('dynflow', "~> 1.0")
+  gem.add_runtime_dependency('dynflow', "~> 1.1")
   gem.add_runtime_dependency('foreman-tasks-core', '>= 0.1.7')
   gem.add_runtime_dependency('sequel')
   gem.add_runtime_dependency('sqlite3')


lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -54,13 +54,28 @@ def roll_log
...
     end
 
+    class ProxyStructuredFormater < ::Dynflow::LoggerAdapters::Formatters::Abstract
+      def call(_severity, _datetime, _prog_name, message)
+        if ::Exception === message
+          @base.exception("Error details", message)
+          "#{message.message} (#{message.class})"
+        else
+          message
+        end
```
lzap: Basically we want to use helper called `exception` which is similar to what we have in Foreman core. It reports exteptions via WARNING level, adds properly formatted stacktrace and it also logs it via structured fields (exception class, message, stacktrace). Then we return the textual form (without joining backtrace).

lzap:
> I’m not sure this behaves exactly as expected. I always thought about the formatter as “let me format the message in any way I like and then just print it”. What happens here is that the formatted message gets processed again before being printed so we get doubled timestamps and stuff like that. We could change the formatter, but it doesn’t feel right.

The purpose of this patch is to remove those extra timestamp actually. This does not look correct, it was working yesterday.

lzap:
Hmmm it works here, how do you configure it again?

lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -54,13 +54,28 @@ def roll_log
...
     class ProxyAdapter < ::Dynflow::LoggerAdapters::Simple
-      def initialize(logger, level = Logger::DEBUG, formatters = [::Dynflow::LoggerAdapters::Formatters::Exception])
+      def initialize(logger, level = Logger::DEBUG, formatters = [])
         @logger           = logger
         @logger.level     = level
-        @logger.formatter = method(:formatter).to_proc
-        @action_logger    = apply_formatters ProgNameWrapper.new(@logger, ' action'), formatters
-        @dynflow_logger   = apply_formatters ProgNameWrapper.new(@logger, 'dynflow'), formatters
+        @logger.formatter = ProxyStructuredFormater.new(@logger)
+        @action_logger    = apply_formatters(ProgNameWrapper.new(@logger, ' action'), [ProxyStructuredFormater])
```
iNecas: When calling `action_logger.info(RuntimeError.new("just a test"))`, it fails with

```
E, [2018-10-23T13:50:43.980012 #21350] ERROR -- /parallel-executor-core/pool default/worker-4: undefined method `exception' for #<Dynflow::LoggerAdapters::Simple::ProgNameWrapper:0x0055ab02321088>
/home/inecas/Projects/ws/foreman-rex/smart_proxy_dynflow/lib/smart_proxy_dynflow_core/log.rb:60:in `call'
/home/inecas/Projects/ws/foreman-rex/smart_proxy_dynflow/lib/smart_proxy_dynflow_core/log.rb:68:in `format'
```
That's probably because the `@base` in this case would not be the `@logger` but the `ProgNameWrapper`. I would need to think of the best way to solve this issue: any suggestions welcome

iNecas: Ok, part of it should be solved by this PR in dynlfow https://github.com/Dynflow/dynflow/pull/300, however, we should still add  `@base.exception("Error details", message) if @base.respond_to?(:exception)` to https://github.com/theforeman/smart_proxy_dynflow/pull/56/files#diff-fe07044a5ec271675cdb2f5043c6ca77R60, as when running smart_proxy_dynflow_core, the logger might not have the `exception` method defined

lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -54,13 +54,28 @@ def roll_log
       end
     end
 
+    class ProxyStructuredFormater < ::Dynflow::LoggerAdapters::Formatters::Abstract
+      def call(_severity, _datetime, _prog_name, message)
+        if ::Exception === message
+          @base.exception("Error details", message)
```
iNecas: To count on the fact that the `exception` method might not be present, it should be IMO something like this:

```ruby
          if @base.respond_to?(:exception)
            @base.exception("Error details", message)
            "#{message.message} (#{message.class})"
          else
            "#{message.message} (#{message.class})\n#{message.backtrace.join("\n")}"
          end
```

lib/smart_proxy_dynflow_core/log.rb
lzap: Ok that works for me. Ideally we would move from Ruby logging stack which is really very simplified and limiting to logging gem in Dynflow with its powerful things like structured logging. Then you could even drop these wrappers completely because you can put things into logging contexts and have them appear in logs.

lib/smart_proxy_dynflow_core/log.rb
lzap: Wait, you just added the method into the wrapper, am I missing something? https://github.com/Dynflow/dynflow/pull/300/files

iNecas: I think we would still miss the `exception` defined in logger when running in separate process, where smart-proxy goodies are not available. Btw. `dynflow-1.1.2` with the required changes as been released: we should probably bump a version requirement here as well.

lzap: Updated with the inline editor via github.com. Not sure I will be able to update the dependency from here, lemme check.

lzap:
Do you mean to update this line? https://github.com/theforeman/smart_proxy_dynflow/blob/master/smart_proxy_dynflow_core.gemspec#L33

iNecas:
Yes

lzap:
Ok updated the dependency.

lib/smart_proxy_dynflow_core/log.rb
iNecas: Still not updated fully, could you add something like this, so that we get the backtrace logged properly even when running outside of smart-proxy?:

```ruby
        if ::Exception === message
          subject = "#{message.message} (#{message.class})"
          if @base.respond_to?(:exception)
            @base.exception("Error details", message)
            subject
          else
            "#{subject}\n#{message.backtrace.join("\n")}"
          end
        else
          message
        end
```

iNecas:
Thanks @lzap

PR #59, opened 2019-03-15
Fixes #26374 - use proxy code for authorization

iNecas:
Also, I've made the core authorization code a bit more defensive:
I've noticed that Puma is setting 'https' in `ENV['HTTPS']`: if we don't
check that we should use https, we would be letting the requests though
even if we shouldn't. Luckily, we're not running Puma just yet, but it's
good to be prepared, so we're actually also checking whether we really
are not using https. Also, just the presence of `ENV['HTTPS']` is now
enough to assume we are in the SSL mode.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -86,7 +86,7 @@ Style/Next:
   Enabled: false
 
 Metrics/AbcSize:
-  Max: 30
+  Enabled: false
 
 Metrics/CyclomaticComplexity:
   Max: 15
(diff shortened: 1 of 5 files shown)


.rubocop.yml
```diff
@@ -86,7 +86,7 @@ Style/Next:
   Enabled: false
 
 Metrics/AbcSize:
-  Max: 30
+  Enabled: false
```
iNecas: Inspired by https://github.com/theforeman/smart-proxy/commit/21642e60255d5ad19e30a97e706ffc7afb6defa4, I don't remember last time the AbcSize metric was of any use for me.

PR #61, opened 2019-04-08
Fixes #26561 - logging stack replaced

lzap:
This replaces the Ruby logging stack with logging gem which provides enhanced features including logging to journald via structured fields.


lzap:
Inspiration at https://github.com/theforeman/smart-proxy/pull/611

This is vaguely tested at the moment, also the code currently does not take advantage of the exception helper method at all and also no correlation id is taken into account. Help me to integrate this - the idea is that a request in Foreman gets UUID which is then passed to Foreman proxy and logged as well. We need to pass this UUID to Dynflow and log it along with all records.

adamruzicka:
Smart proxy proxies requests to smart proxy dynflow core. The context needs to be extracted from the incoming request and passed into the outgoing request somewhere around here https://github.com/theforeman/smart_proxy_dynflow/blob/master/lib/smart_proxy_dynflow/callback.rb#L22 .

This context the needs to be read from the incoming request in smart proxy dynflow core's API https://github.com/theforeman/smart_proxy_dynflow/blob/master/lib/smart_proxy_dynflow_core/api.rb , be used for sinatra's logging and then passed down into the action which is being triggered. 

Then we will need to introduce some kind of middleware to set/unset the correct logging context when an action is being run. Something like https://github.com/theforeman/foreman-tasks/pull/383 should do the trick.

@inecas did I miss anything?

iNecas:
If we would be ok with the API calls on smart-proxy-dynflow and smart-proxy-dynflow-core not using the correlation id, but just using it in the core dynflow tasks, the mechanism might be a bit simpler, as it would mean just passing it as input for the proxy tasks and restoring the mdc via middleware.

lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -67,20 +110,32 @@ def call(_severity, _datetime, _prog_name, message)
...
         @logger.level     = level
-        @logger.formatter = ProxyStructuredFormater.new(@logger)
         @action_logger    = apply_formatters(ProgNameWrapper.new(@logger, ' action'), [ProxyStructuredFormater])
         @dynflow_logger   = apply_formatters(ProgNameWrapper.new(@logger, 'dynflow'), [ProxyStructuredFormater])
       end
     end
   end
 end
+
+class PryAction < ::Dynflow::Action
```
adamruzicka: This should go away before merge

lzap:
If I want to test this though proxy, what is the URL? The curl above is giving me 404 (spd is enabled and initialized).

lzap:
Ok I think this is now ready for review and testing. This is how it works, smart proxy log:

```
2019-04-12T13:11:08  [D] accept: 192.168.1.13:42116
2019-04-12T13:11:08  [D] Rack::Handler::WEBrick is invoked.
2019-04-12T13:11:08 9e428de6 [I] Started GET /console 
2019-04-12T13:11:08 9e428de6 [D] Headers: {"HTTP_HOST"=>"box.home.lan:8448", "HTTP_USER_AGENT"=>"curl/7.61.1", "HTTP_ACCEPT"=>"*/*", "HTTP_VERSION"=>"HTTP/1.1"}
2019-04-12T13:11:08 9e428de6 [D] require_ssl_client_verification: skipping, non-HTTPS request
2019-04-12T13:11:08 9e428de6 [D] verifying remote client box (based on REMOTE_ADDR) against trusted_hosts ["box", "box.home.lan", "localhost", "127.0.0.1", "192.168.122.1", "foreman.local.lan", "foreman.provision.lan"]
2019-04-12T13:11:08 9e428de6 [D] Proxy request from box.home.lan:8448/dynflow/console to http://127.0.0.1:8008/console
2019-04-12T13:11:08 9e428de6 [D] Proxy request status 200 - #<Net::HTTPOK:0x0000555557e3a1a0>
2019-04-12T13:11:08 9e428de6 [I] Finished GET /console with 200 (9.07 ms)
2019-04-12T13:11:08  [D] close: 192.168.1.13:42116
^C2019-04-12T13:11:16  [D] close TCPSocket(0.0.0.0, 8443)
```

Those debug messages without any request id (`9e428de6`) are coming from webrick, we have it enabled (hardcoded). For dynflow core I followed the very same approach as in proxy: the rack common and access loggers are both disabled and there is a custom middleware which logs requests along with extra debugging information into the debug mode when enabled:

```
2019-04-12T13:11:08  [D] accept: 127.0.0.1:39922
2019-04-12T13:11:08  [D] Rack::Handler::WEBrick is invoked.
2019-04-12T13:11:08 9e428de6 [I] Started GET  
2019-04-12T13:11:08 9e428de6 [D] Headers: {"HTTP_ACCEPT_ENCODING"=>"gzip;q=1.0,deflate;q=0.6,identity;q=0.3", "HTTP_ACCEPT"=>"*/*, application/json,version=2", "HTTP_USER_AGENT"=>"Ruby", "HTTP_X_FORWARDED_FOR"=>"box.home.lan:8448", "HTTP_X_REQUEST_ID"=>"9e428de6-b6d1-47b0-8546-f4f364d6d4f6", "HTTP_CONNECTION"=>"close", "HTTP_HOST"=>"127.0.0.1:8008", "HTTP_VERSION"=>"HTTP/1.1"}
2019-04-12T13:11:08 9e428de6 [D] 
2019-04-12T13:11:08 9e428de6 [I] Finished GET  with 200 (4.43 ms)
2019-04-12T13:11:08  [D] close: 127.0.0.1:39922
```

As you can see the the request ID is passed into the core and logged along with other structured info.

However, there is still one issue described above - dynflow log lines do not contain the request id until the method `restore_current_request_id` of the provided middleware is called. I see no way of inserting dynflow middleware to the very first position so it's called early before anything else happens. Any idea @iNecas ? Do we need to implement it? I guess so.

lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -1,61 +1,106 @@
...
+      def reopen
+        return if @logger.nil?
+        @logger.appenders.each(&:reopen)
+      end
+
+      def instance
+        return @logger if @logger
+        layout = Logging::Layouts.pattern(pattern: setting(:file_logging_pattern, '%d %.8X{request} [%.1l] %m') + "\n")
+        notime_layout = Logging::Layouts.pattern(pattern: setting(:system_logging_pattern, '%.8X{request} [%.1l] %m') + "\n")
+        log_file = setting(:log_file, 'STDOUT')
```
iNecas: The defaults are being set here https://github.com/theforeman/smart_proxy_dynflow/blob/37e9a46f13f3d1b572a41ea471713fa13b631dbf/lib/smart_proxy_dynflow_core/settings.rb#L38

lib/smart_proxy_dynflow_core/log.rb
iNecas: Or we could turn it into something like `SmartProxyDynflowCore::DummyAction` with docs comments on how to use it: it might be quite nice debugging setup.

lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -67,20 +112,32 @@ def call(_severity, _datetime, _prog_name, message)
...
+class PryAction < ::Dynflow::Action
+  require 'pry'
+
+  def plan(args)
+    # logging
+    plan_self
+  end
+
+  def run
+    #binding.pry
```
iNecas: We can teach it, how to raise an error based on action_input
```ruby
raise "error" if input[:raise_exception]
```

lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -67,20 +112,32 @@ def call(_severity, _datetime, _prog_name, message)
...
         @dynflow_logger   = apply_formatters(ProgNameWrapper.new(@logger, 'dynflow'), [ProxyStructuredFormater])
       end
     end
   end
 end
+
+class PryAction < ::Dynflow::Action
+  require 'pry'
+
+  def plan(args)
```
iNecas: This method should not be needed: the default one just expects the `args` to be hash and passes it as action input

lib/smart_proxy_dynflow_core/log.rb
adamruzicka: Shouldn't this then live directly in Dynflow so it can be used anywhere?

lib/smart_proxy_dynflow_core/logger_middleware.rb
```diff
@@ -0,0 +1,28 @@
...
+      env['rack.input'].try(:seek, 0) if @logger.debug?
+      before = Time.now.to_f
+      status, _, _ = @app.call(env)
+    rescue Exception => e
+      @logger.exception "Error processing request '#{::Logging.mdc['request']}", e
+      raise e
+    ensure
+      @logger.info do
+        after = Time.now.to_f
+        duration = (after - before) * 1000
```
iNecas: we could check whether the `before` is initialized, in case the exception is raised before line 15

lib/smart_proxy_dynflow_core/logger_middleware.rb
```diff
@@ -0,0 +1,28 @@
...
+      @app = app
+    end
+
+    def call(env)
+      status = 500
+      env['rack.logger'] = @logger
+      @logger.info { "Started #{env['REQUEST_METHOD']} #{env['PATH_INFO']} #{env['QUERY_STRING']}" }
+      @logger.debug { 'Headers: ' + env.select {|k,v| k.start_with? 'HTTP_'}.inspect }
+      @logger.debug { (body = env['rack.input'].read).empty? ? '' : 'Body: ' + body }
+      env['rack.input'].try(:seek, 0) if @logger.debug?
```
iNecas: would be better to put this inot the block above + I would prefer `rewind` instead

lzap:
So I tested this with STDOUT, JOURNAL and SYSLOG. All works fine, fixed all comments. Here is an example how an entry looks like, not many custom fields except REQUEST id and REMOTE_IP and SYSLOG_IDENTIFIER and PRIORITY (= logging level) for now, but it's a good start.

```
# journalctl -f -o verbose
    _SOURCE_REALTIME_TIMESTAMP=1555573546705597
Thu 2019-04-18 09:45:46.707400 CEST [s=927adf6bc1964d178dc5ac5cf1f5a58b;i=25c15c;b=44de32f5987c4f1b86e415cad1cfbd5d;m=1168f946b;t=586c930ac69d3;x=14446913b24dc7d2]
    PRIORITY=6
    _TRANSPORT=journal
    _UID=1000
    _GID=1000
    _CAP_EFFECTIVE=0
    _AUDIT_LOGINUID=1000
    _SYSTEMD_OWNER_UID=1000
    _SYSTEMD_SLICE=user-1000.slice
    _SYSTEMD_USER_SLICE=-.slice
    _MACHINE_ID=2a2fb6fa375d459da95c6ada4a67d03a
    _HOSTNAME=box.home.lan
    _AUDIT_SESSION=1
    _SYSTEMD_CGROUP=/user.slice/user-1000.slice/session-1.scope
    _SYSTEMD_SESSION=1
    _SYSTEMD_UNIT=session-1.scope
    _COMM=bundle
    _EXE=/home/lzap/.rbenv/versions/2.5.1/bin/ruby
    _BOOT_ID=44de32f5987c4f1b86e415cad1cfbd5d
    _SYSTEMD_INVOCATION_ID=5d6c9179b23f48489112e16431fb9b6f
    SYSLOG_IDENTIFIER=dynflow-core
    SYSLOG_FACILITY=168
    _PID=31062
    _CMDLINE=bin/smart_proxy_dynflow_core
    REMOTE_IP=127.0.0.1
    REQUEST=f0992764-2d4d-43dd-8e4c-061e82b263d0
    MESSAGE=f0992764 [I] Finished POST /tasks/ with 200 (16.04 ms)
```

I don't have rsyslog on my Fedora, however I tested this with journald which listens for syslog events as well. When testing this, note that there are no custom fields as syslog protocol is only PRIORITY + string message. That's the limitation of it.

lzap:
Fixed rubocop issues

iNecas:
This is needed to fix the tests https://github.com/lzap/smart_proxy_dynflow/pull/2

lzap:
This is a terrible test honestly:

```

  1) Failure:
SmartProxyDynflowCore::HelpersTest#test_0003_skips client cert authentication if token succeeds [/home/jenkins/workspace/smart-proxy-dynflow-pull-request/ruby/2.3/lib/smart_proxy_dynflow_core/logger_middleware.rb:13]:
unexpected invocation: #<Logging::Logger:0x424d3e0>.debug()
unsatisfied expectations:
- expected exactly once, not yet invoked: #<Logging::Logger:0x424d3e0>.debug("authorized with token")
```

;-)

iNecas:
@lzap are you going to merge https://github.com/lzap/smart_proxy_dynflow/pull/2, fix it otherwise, or just keep ranting about terrible tests? :)

lib/smart_proxy_dynflow_core/launcher.rb
```diff
@@ -47,7 +49,7 @@ def self.route_mapping(rack_builder)
 
     def install_usr1_trap
       trap(:USR1) do
-        Log.instance.roll_log
+        Log.reopen
```
iNecas: Perhaps, instead of complaining about quality of old code, you might want to spend the time testing the stuff a bit more so that this one is top class :) This currently gives:

```
# pkill -f -USR1 bin/smart_proxy_dynflow_core
 2019-04-24T16:53:11  [E] ThreadError: can't be called from trap context
	/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/logging-2.2.2/lib/logging/utils.rb:159:in `synchronize'
```

Seems like there is a reason for the code in smart-proxy not being that straight forward https://github.com/theforeman/smart-proxy/blob/8652a7e5c63009cae026f6350178583dcadc5f34/lib/proxy/log_buffer/decorator.rb#L25

lzap:
> @lzap are you going to merge [lzap#2](https://github.com/lzap/smart_proxy_dynflow/pull/2), fix it otherwise, or just keep ranting about terrible tests? :)

Your patch is still poor code, why I think it's important is that users can disable debug level for tests (I do it quite often and only enable it when debugging something weird). The change can break behavior of tests. And I was not just complaining, I was going to fix it. But thanks, let's at least move forward with this. I am going to incorporate your PR into my one. :-)

lib/smart_proxy_dynflow_core/launcher.rb
lzap: Nice, thanks for the insight. Amended the very same approach.

ares:
What is the status here?

iNecas:
/me needs to re-test: haven't noticed the log rotation has been addressed already

lib/smart_proxy_dynflow_core/log.rb
```diff
@@ -1,61 +1,112 @@
...
         end
       end
 
-      def log_file
-        if Settings.instance.loaded && Settings.instance.log_file
-          Settings.instance.log_file
+      def handle_reopen
+        @logger.warn "Performing log file reopen"
+        @reopen = false
+        @logger.appenders.each(&:reopen)
```
iNecas: There still seems to be some flaw there: when I try:

1. `mv logs/smart_proxy_dynflow_core.log{,.bak}`
2. `pkill -f -USR1 bin/smart_proxy_dynflow_core`
3. `curl localhost:8008/tasks/ -X POST -d '{ "action_name": "SmartProxyDynflowCore::TestAction", "action_input": {"exception": "false"} }'`
4.  `du logs/smart_proxy_dynflow_core.log`:
```
0	logs/smart_proxy_dynflow_core.log
```

So the new file is re-created, but no output is getting in. Seems like there is some additional logic in smart-proxy that actually makes it work (the smart-proxy worked fine when trying this)

lzap: Good find, yeah the issue is that I call `handle_reopen` only in `instance` which is obviously only called during initialization. I wanted to avoid overloading logging methods but it looks like it's the only way.

lzap:
There were two problems, added new commit for easier review:

1) As I described, reopen flag was set correctly but only picked up when `instance` method was called. I created a dummy appender with only job to initiate reopen when new message arrives and the flag is set.

2) You won't see new messages appearing in the reopened file because of buffering. This revealed actually a bug - the process did not close logging after exit. Now it logs an INFO log line "Finished shutting down" and then closes all logging. So to test just exit the process to see the result or send enough `curl` command for the buffer to flush.

iNecas:
It's still strange that the log works fine (I see the messages as I do the request) until the first reopen. I actually still don't see the messages, even on exit. I'm afraid there is still some bug in there

iNecas:
How about something like this:

```patch
diff --git a/lib/smart_proxy_dynflow_core/launcher.rb b/lib/smart_proxy_dynflow_core/launcher.rb
index c099198..18007e8 100644
--- a/lib/smart_proxy_dynflow_core/launcher.rb
+++ b/lib/smart_proxy_dynflow_core/launcher.rb
@@ -14,6 +14,7 @@ module SmartProxyDynflowCore
       load_settings!(options)
       Settings.instance.standalone = true
       install_usr1_trap
+      require 'pry'; binding.pry
       Rack::Server.new(rack_settings).start do |_server|
         SmartProxyDynflowCore::Core.ensure_initialized
       end
diff --git a/lib/smart_proxy_dynflow_core/log.rb b/lib/smart_proxy_dynflow_core/log.rb
index dc23be6..1c18fd6 100644
--- a/lib/smart_proxy_dynflow_core/log.rb
+++ b/lib/smart_proxy_dynflow_core/log.rb
@@ -36,6 +36,7 @@ module SmartProxyDynflowCore
         Logging.logger[LOGGER_NAME].appenders.each(&:close)
         Logging.logger[LOGGER_NAME].clear_appenders
         @logger = nil
+        @do_reopen = false
         instance
       end
 
@@ -43,13 +44,15 @@ module SmartProxyDynflowCore
         return if @logger.nil?
         if Settings.instance.log_file !~ /^(STDOUT|SYSLOG|JOURNALD?)$/i
           @logger.warn "Scheduled log file reopen"
-          @reopen.set
+          @do_reopen = true
+          #@reopen.set
         else
           @logger.debug "Not scheduling log file reopen (no file appender)"
         end
       end
 
       def instance
+        reload! if @do_reopen
         return @logger if @logger
         layout = Logging::Layouts.pattern(pattern: Settings.instance.file_logging_pattern + "\n")
         notime_layout = Logging::Layouts.pattern(pattern: Settings.instance.system_logging_pattern + "\n")
```

The reloading should be probably covered with mutex, to ensure we do it only only once.

lzap:
You say after you exit you do not see the final message? Can you show me repro steps?

The solution you suggest will not work since `instance` is only called once and memorized. I agree with the mutex tho.

lib/smart_proxy_dynflow_core/launcher.rb
adamruzicka: Do we even need the usr1 trap? Originally this was here because the log rotation was done by an external process. If I understand this correctly log rotation will be handled by the process itself so it doesn't need to be notified from the outside. Or am I missing something?

adamruzicka:
I think I managed to reproduce the issue @iNecas mentions.

Steps:
1) Apply the patch from this gist https://gist.github.com/adamruzicka/d0c69b0e716f6f9b1b78c692278269b7
2) Set smart proxy dynflow to log to a file and to rotate it after it grows above 1MB
3) Trigger a bunch of counter actions
```shell
for i in `seq 100`; do
  curl localhost:8008/tasks/ -X POST -d '{ "action_name": "SmartProxyDynflowCore::CounterAction", "action_input": { "target": 10000000 } }'
  sleep 0.5
done
```

After a rotation the log file will get reopened, but not more logs will appear there. To me it looks like some kind of a race condition

lib/smart_proxy_dynflow_core/middleware/keep_current_request_id.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      def store_current_request_id
+        action.input[:current_request_id] = ::Logging.mdc['request']
+      end
+
+      def restore_current_request_id
+        unless (restored_id = action.input[:current_request_id]).nil?
+          old_id = ::Logging.mdc['request']
+          if !old_id.nil? && old_id != restored_id
+            action.action_logger.warn(_('Changing request id %{request_id} to saved id %{saved_id}') % { :saved_id => restored_id, :request_id => old_id })
```
adamruzicka: `undefined method '_' for #<Actions::Middleware::KeepCurrentRequestID:0x000056547ca803a0> (NoMethodError)`

lzap:
I rebased, added mutex around reopen flag although I think it has no effect. It works for me, this is how you can test it:

```yaml
:log_file: /tmp/smart_proxy_dynflow_core.log
:log_level: DEBUG
:file_rolling_size: 1
```

And then watch md5 sums of the key files:

watch -n0.5 md5sum /tmp/smart_proxy_dynflow_core*

For me it is changing over and over again, the rotated file is being correctly overwritten. Note that logging library does not create new and new files as the default rotate policy ("roll_by") is date-based. This is subjective preference by me as I prefer data-based rotation. If you set file_rolling_age to some arbitrary number like 10 (seconds) then logging will also append time to rotated file, then you will start seeing more expected files.

```
2323472646289ef2f7ca20e33b315824  /tmp/smart_proxy_dynflow_core.log
d41d8cd98f00b204e9800998ecf8427e  /tmp/smart_proxy_dynflow_core.log.age
a0a545a939db52b15d3691c001bc8860  /tmp/smart_proxy_dynflow_core.20190510.log
```

lzap:
I have to admit this is confusing and we might decide to change this behavior to just numbered files like `xxx.1.log, xxx.2.log` etc. I'd change this in proxy too because it also uses this new naming scheme now too.

adamruzicka:
Even with the mutex in place I can reproduce the issue

lzap:
Ok it's a bug in the logging gem: https://github.com/TwP/logging/pull/206

lib/smart_proxy_dynflow_core/launcher.rb
lzap: No it was not decided yet and I am actually leaning towards keeping logrotate after some more thoughts. This is here only because logging gem supports it so there is an option e.g. for gem installations. However I want to make rolling optional now.

lzap:
Ok the reason for the rolling problem was in logging gem, with that I think the most reasonable approach is to default to normal file logger and rely on logrotate for both dynflow and smart proxy. I changed the :file_rolling_size: setting behavior - when set to 0 rolling is disabled which is now the default. The bug is still there until we upgrade to fixed logging gem but this is opt in and we will be using logrotate by default.

I will go ahead and once this is merged do the same thing for smart-proxy. For Foreman core we do use logrotate only so no change there. This is now ready for re-review and re-testing @iNecas and @adamruzicka

lzap:
Take care when testing this, file appender is like 2-3x faster than the rolling appender (when rolling often) so it might look like it gets stuck but it actually finishes quicker.

lzap:
I now understand the reason why the logging stack stops writing, it's because in the case of exception it sets the global level to `:off`. I will come back later to fix this as I have a different priority now. Thanks for now.

adamruzicka:
@lzap any chance this could get moving again in the foreseeable future?

lzap:
Depends on what we want. Currently, auto log rotating does not work in multithreaded environment correctly. By default, we use logrotate for production, so it's only issue for development environments. Several options: we can wait if TwP fixes the problem. Or we can remove file-logging feature completely from dynflow (and smart proxy and foreman?).

I poked the maintainer once again to see if there are any plans short-term, let's see.

lzap:
Upstream released, packaging PR: https://github.com/theforeman/foreman-packaging/pull/5783

lzap:
While testing I am running into this when I do Ctrl-C:

```
bundler: failed to load command: bin/smart_proxy_dynflow_core (bin/smart_proxy_dynflow_core)
NoMethodError: undefined method `shutdown' for nil:NilClass
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/rack-2.0.5/lib/rack/handler/webrick.rb:48:in `shutdown'
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/rack-2.0.5/lib/rack/server.rb:291:in `block in start'
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/2.7.0/webrick/server.rb:371:in `close'
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/2.7.0/webrick/server.rb:371:in `block in cleanup_listener'
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/2.7.0/webrick/server.rb:357:in `each'
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/2.7.0/webrick/server.rb:357:in `cleanup_listener'
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/2.7.0/webrick/server.rb:205:in `block in start'
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/2.7.0/webrick/server.rb:32:in `start'
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/2.7.0/webrick/server.rb:157:in `start'
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/rack-2.0.5/lib/rack/handler/webrick.rb:34:in `run'
  /home/lzap/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/rack-2.0.5/lib/rack/server.rb:297:in `start'
  /home/lzap/work/smart_proxy_dynflow/lib/smart_proxy_dynflow_core/launcher.rb:22:in `start'
  /home/lzap/work/smart_proxy_dynflow/lib/smart_proxy_dynflow_core/launcher.rb:15:in `launch!'
  bin/smart_proxy_dynflow_core:32:in `<top (required)>'
```

Is this a known issue?

lzap:
Anyway, I just repoduced this with 2.2.2 gem, then I updated to the new 2.3.0 release and `pkill` no longer triggers the problem. This is now ready for re-review. (No changes there except merging ton of conflicts)

adamruzicka (commented):
When running as external core I'm getting a blank line out of nowhere in the logs
```
2020-09-10T13:42:06 795a45de [I] Started GET /tasks/count 
2020-09-10T13:42:06 795a45de [D] Headers: {"HTTP_ACCEPT_ENCODING"=>"gzip;q=1.0,deflate;q=0.6,identity;q=0.3", "HTTP_ACCEPT"=>"*/*, application/json,version=2", "HTTP_USER_AGENT"=>"Ruby", "HTTP_X_FORWARDED_FOR"=>"localhost:8000", "HTTP_X_REQUEST_ID"=>"795a45de-82fd-4813-b103-d50821352e33", "HTTP_CONNECTION"=>"close"
, "HTTP_HOST"=>"127.0.0.1:8008", "HTTP_VERSION"=>"HTTP/1.1"}
2020-09-10T13:42:06 795a45de [D] 
2020-09-10T13:42:06 795a45de [D] require_ssl_client_verification: skipping, non-HTTPS request
2020-09-10T13:42:06 795a45de [I] Finished GET /tasks/count with 200 (1.05 ms)
2020-09-10T13:42:06  [D] close: 127.0.0.1:33212
```

If I run with dynflow core inside smart proxy (and external_core set to no, the debian way), I'm getting doubled outputs. Also note two request ids even though there was only one request
```
2020-09-10T13:43:24 5e3f7c2b [I] Started GET /tasks/count 
2020-09-10T13:43:24 5e3f7c2b [I] Started GET /tasks/count 
2020-09-10T13:43:24 5e3f7c2b [D] Headers: {"HTTP_HOST"=>"localhost:8000", "HTTP_USER_AGENT"=>"curl/7.66.0", "HTTP_ACCEPT"=>"*/*", "HTTP_VERSION"=>"HTTP/1.1"}
2020-09-10T13:43:24 5e3f7c2b [D] Headers: {"HTTP_HOST"=>"localhost:8000", "HTTP_USER_AGENT"=>"curl/7.66.0", "HTTP_ACCEPT"=>"*/*", "HTTP_VERSION"=>"HTTP/1.1"}
2020-09-10T13:43:24 5e3f7c2b [D] 
2020-09-10T13:43:24 5e3f7c2b [D] 
2020-09-10T13:43:24 4d345aae [I] Started GET /dynflow/tasks/count 
2020-09-10T13:43:24 4d345aae [D] require_ssl_client_verification: skipping, non-HTTPS request
2020-09-10T13:43:24 4d345aae [D] require_ssl_client_verification: skipping, non-HTTPS request
2020-09-10T13:43:24 4d345aae [I] Finished GET /dynflow/tasks/count with 200 (3.64 ms)
2020-09-10T13:43:24 4d345aae [I] Finished GET /tasks/count with 200 (4.25 ms)
2020-09-10T13:43:24 4d345aae [I] Finished GET /tasks/count with 200 (4.25 ms)
```

config/settings.yml.example
```diff
@@ -39,11 +39,27 @@
 # Specify versions like: '1.1', or '1.2'
 #:tls_disabled_versions: []
 
-# File to log to, leave empty for logging to STDOUT
-# :log_file: /var/log/foreman-proxy/smart_proxy_dynflow_core.log
+# File to log to, leave empty for stdout or use STDOUT, STDERR, SYSLOG or JOURNAL
```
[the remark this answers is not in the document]
lzap: Ok resolving.

lzap:
Double lines problem solved, the middleware must not be included when running in standalone mode. Packaging PR still not merged but this should be ready: https://github.com/theforeman/foreman-packaging/pull/5666

lzap:
Dependency merged into nightly for the record.

adamruzicka (commented):
In general it works well, left some nitpicks in the code

lib/smart_proxy_dynflow_core/test_action.rb
```diff
@@ -0,0 +1,21 @@
...
+    def run
+      state = 0
+      while state < input[:debug_logs]
+        action_logger.debug("Test action counter at #{state}")
+        state += 1
+      end
+      raise "test exception" if input[:exception]
+    end
+  end
+end
```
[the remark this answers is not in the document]
lzap: Uh I am not aware writing this or adding this hunk. Perhaps bad conflict resolution? I can delete this for sure if you think it's not valid.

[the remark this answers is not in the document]
lzap: Oh now I remember :-)

Gemfile
```diff
@@ -32,6 +32,7 @@ else
   gem 'rack', '>= 1.1'
   gem 'sinatra'
 end
+gem 'logging-journald', '~> 2.0', :platforms => [:ruby]
```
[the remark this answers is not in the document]
lzap: I could not find a way how to add a gemspec dependency only for a particular platform. This one won't work on JRuby or Ruby on Windows. Anyway, I am going to follow what smart-proxy does: https://github.com/theforeman/smart-proxy/blob/develop/smart_proxy.gemspec

Adding runtime dependency "logging" and keeping this line here as well.

lib/smart_proxy_dynflow_core/logger_middleware.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def call(env)
+      before = Time.now.to_f
+      status = 500
+      env['rack.logger'] = @logger
+      @logger.info { "Started #{env['REQUEST_METHOD']} #{env['PATH_INFO']} #{env['QUERY_STRING']}" }
+      @logger.debug { 'Headers: ' + env.select { |k, v| k.start_with? 'HTTP_' }.inspect }
+      if @logger.debug? && env['rack.input']
+        body = env['rack.input'].read
+        @logger.debug body.empty? ? '' : 'Body: ' + body
```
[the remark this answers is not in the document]
lzap: Right, good point.

lzap:
Rebased.

lzap:
We should be all good @adamruzicka ? Or did I miss some comment(s)?

adamruzicka:
This should be fine, I was running with this for the past week or so. I just need to read it through once more and it will go in

adamruzicka (changes requested):
One last thing

adamruzicka (changes requested):
Just drop the last offending require and then we'll get it in. And apparently it also needs a rebase

lzap:
Rebased, I had a bunch of conflicts in gemspec/gemfiles so please re-review carefully.

adamruzicka:
:police_officer: is not happy

lzap:
Well.

```
lib/smart_proxy_dynflow_core/log.rb:78:16: C: Style/NumericPredicate: Use size.positive? instead of size > 0.
            if size > 0
```

Ruby/Rubocop community is just swimming against current. It's `size > 0` in every major language out there... nonsense. Fixed, my hate against Rubocop just increased a bit once again.

adamruzicka (approved):
ACK

adamruzicka:
Thank you @lzap !

PR #62, opened 2019-04-26
Fixes #26691 - add `POST /tasks/:task_id/update` endpoint

iNecas:
Propagates external events to underlying action without invalidating the
OTP, so that it can be called multiple times.

Depends on:

- [x] - https://github.com/theforeman/foreman-tasks/pull/416

### DIFF
--- a/lib/smart_proxy_dynflow_core/api.rb
+++ b/lib/smart_proxy_dynflow_core/api.rb
@@ -6,7 +6,13 @@ module SmartProxyDynflowCore
     helpers Helpers
 
     before do
-      authorize_with_token || authorize_with_ssl_client
+      if request.path_info =~ %r{/tasks/.*/done}
+        authorize_with_token(clear: true)
+      elsif request.path_info =~ %r{/tasks/.*/update}
+        authorize_with_token(clear: false)
+      else
+        authorize_with_ssl_client
+      end
       content_type :json
     end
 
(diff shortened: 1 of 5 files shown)


lib/smart_proxy_dynflow_core/api.rb
```diff
@@ -45,9 +51,14 @@ class Api < ::Sinatra::Base
       tasks_count(params['state']).to_json
     end
 
+    post "/tasks/:task_id/update" do |task_id|
```
adamruzicka: How about matching the route with a regexp and providing a single body since the behavior of `/update` and `/done` is the same?

adamruzicka:
[test]

adamruzicka:
[test smart_proxy_dynflow]

lib/smart_proxy_dynflow_core/api.rb
iNecas: Updated

iNecas:
The tests are currently expected to fail due to missing tasks dependency change, but when running locally, some tests were failing so I've fixed that (see https://github.com/theforeman/smart_proxy_dynflow/pull/62/commits/1b4d17da932fcda7ddff32eb34435f89edbf528f). This required additional commit to the tasks one (https://github.com/theforeman/foreman-tasks/pull/416/commits/49b81d9ccd8943c6816d6bbe1ab1378124a7abb7) to accept the expected_user param.

adamruzicka:
Thank you @iNecas !

PR #63, opened 2019-06-10
Fixes #26969 - Allow using remote core even if core gem is available

adamruzicka:
The behavior can be configured using the `external_core` key in the settings.

If unset, the behavior will remain the same, eg. use internal core if available,
fallback to external core otherwise. Setting it to `true` will force the proxy to
always use an external core. Setting it to `false` will make the proxy try to
always use the internal core and disable the feature if it is not available.

### DIFF
--- a/lib/smart_proxy_dynflow/plugin.rb
+++ b/lib/smart_proxy_dynflow/plugin.rb
@@ -4,7 +4,7 @@ require 'proxy/plugin'
 
 class Proxy::Dynflow
   class Plugin < Proxy::Plugin
-    rackup_path = File.expand_path('http_config.ru', File.expand_path("../", __FILE__))
+    rackup_path = File.expand_path('http_config.ru', __dir__)
     http_rackup_path rackup_path
     https_rackup_path rackup_path
 
--- a/lib/smart_proxy_dynflow/plugin.rb
+++ b/lib/smart_proxy_dynflow/plugin.rb
@@ -4,29 +4,18 @@ require 'proxy/plugin'
 
 class Proxy::Dynflow
   class Plugin < Proxy::Plugin
-    rackup_path = begin
-      require 'smart_proxy_dynflow_core'
-      'http_config_with_executor.ru'
-    rescue LoadError
-      'http_config.ru'
-    end
-    http_rackup_path File.expand_path(rackup_path, File.expand_path("../", __FILE__))
-    https_rackup_path File.expand_path(rackup_path, File.expand_path("../", __FILE__))
+    rackup_path = File.expand_path('http_config.ru', File.expand_path("../", __FILE__))
+    http_rackup_path rackup_path
+    https_rackup_path rackup_path
 
     settings_file "dynflow.yml"
     requires :foreman_proxy, ">= 1.12.0"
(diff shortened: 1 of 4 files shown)


lib/smart_proxy_dynflow/plugin.rb
```diff
@@ -4,29 +4,18 @@
...
-    # rubocop:disable Lint/HandleExceptions
     after_activation do
-      begin
-        require 'smart_proxy_dynflow_core'
-      rescue LoadError
-        # Dynflow core is not available in the proxy, will be handled
-        # by standalone Dynflow core
-      end
+      # Ensure the core gem is loaded, if configure NOT to use the external core
+      require 'smart_proxy_dynflow_core' if Proxy::Dynflow::Plugin.settings.external_core == false
```
ekohl:
```suggestion
      require 'smart_proxy_dynflow_core' unless Proxy::Dynflow::Plugin.settings.external_core
```

lib/smart_proxy_dynflow/plugin.rb
```diff
@@ -4,29 +4,18 @@
...
   class Plugin < Proxy::Plugin
-    rackup_path = begin
-      require 'smart_proxy_dynflow_core'
-      'http_config_with_executor.ru'
-    rescue LoadError
-      'http_config.ru'
-    end
-    http_rackup_path File.expand_path(rackup_path, File.expand_path("../", __FILE__))
-    https_rackup_path File.expand_path(rackup_path, File.expand_path("../", __FILE__))
+    rackup_path = File.expand_path('http_config.ru', File.expand_path("../", __FILE__))
```
ekohl: Isn't this equvalent to:
```suggestion
    rackup_path = File.expand_path('http_config.ru', __dir__)
```

lib/smart_proxy_dynflow/plugin.rb
adamruzicka: The thing is I need to distinguish between `false` and `nil`. But maybe I'm just overthinking things and this whole three-valued logic is not needed at all

lib/smart_proxy_dynflow/plugin.rb
adamruzicka: Good point, this is a remnant of supporting ruby `< 2.0`. Will change

lib/smart_proxy_dynflow/plugin.rb
ekohl: Since there's duplication in the http_config, should we have a helper? Can we reuse the output of this loading somehow?

iNecas: Btw. would this work without Debian packaging/installer changes to explicitly disable the loading?

adamruzicka: Yes, by default the value will be unset. This will make the proxy try loading the core, use it if available and fallback to external core otherwise.

iNecas:
smart_proxy_dynflow 0.2.3 released to rubygems and packaging PRs are https://github.com/theforeman/foreman-packaging/pull/3849 and https://github.com/theforeman/foreman-packaging/pull/3850

PR #64, opened 2019-06-28
Fixes #27184 - Add systemd notify support

adamruzicka:


adamruzicka:
@ekohl opinions on the approach taken here?

ekohl:
In itself :+1: but you can take it a step further as the Foreman Proxy did and implement `type=notify`. https://github.com/theforeman/smart-proxy/commit/99e9e5bf5843ea1f05c62febc24bd0da919adc98 is how that was done.

The difference is that with `type=simple` it will start up and can still fail. With `type=notify` it will only be considered started if it successfully bound to the port(s).

adamruzicka:
Huh, I thought that notify required talking to systemd over dbus.

ekohl:
No, not dbus. It receives the a unix socket path via an environment variable.

adamruzicka:
`s/dbus/some socket` :) I'll see what I can do about it

lzap:
Merged.

lib/smart_proxy_dynflow_core/sd_notify.rb
```diff
@@ -0,0 +1,33 @@
+# Shamelessly taken from theforeman/smart-proxy @ 99e9e5b
+# kudos to domcleal
+
+require 'socket'
+
+# Implementation of libsystemd's sd_notify API, sends current state via socket
+module SmartProxyDynflowCore
```
ekohl, line 7: This makes me wonder: there's a lot of duplication between smart-proxy and smart_proxy_dynflow_core to set up the daemon and listeners. I'm wondering if we can unify that somehow. Food for thought over a beer some time.

PR #65, opened 2019-09-17
Fixes #27865 - previous definition and unused CBC ciphers

lzap:
Dynflow plugin has a copy of file from core which issues warnings during startup:

    /home/lzap/work/smart_proxy_dynflow/lib/smart_proxy_dynflow_core/webrick-patch.rb:3: warning: already initialized constant CIPHERS
    /home/lzap/work/smart-proxy/lib/webrick-patch.rb:3: warning: previous definition of CIPHERS was here


adamruzicka:
We need the patch in case smart proxy dynflow runs as a standalone process, for example on production deployments on EL*. How about selectively requiring the patch only if smart proxy dynflow does not run inside regular smart proxy?

lzap:
Oh, can you give me quick hint how to do that? :-)

lib/smart_proxy_dynflow_core/launcher.rb
```diff
@@ -1,7 +1,7 @@
 require 'webrick/https'
 require 'smart_proxy_dynflow_core/bundler_helper'
 require 'smart_proxy_dynflow_core/settings'
-require 'smart_proxy_dynflow_core/webrick-patch'
```
adamruzicka: Something like the following snippet should do the trick
```ruby
require 'smart_proxy_dynflow_core/webrick-patch' unless defined?(::Proxy::Launcher)
```

We can't check if just `::Proxy` is defined because `smart_proxy_dynflow` openes that module at it will always be present

lzap: Oh just like that, ok. Rebased.

lzap:
I've rebased the patch to reflect a core change and renamed the issue and PR as well:

https://github.com/theforeman/smart-proxy/pull/684

lzap:
Made it happy then.

adamruzicka:
Thanks @lzap !

PR #66, opened 2019-09-30
Update to bundler.d to match other smart proxy plugins

ehelms:


ehelms:
Do I understand correctly that at runtime, smart_proxy_dynflow_core looks in `bundler.d` (similar to Foreman and the Smart Proxy) for any plugins that are enabled to run within the context of smart_proxy_dynflow_core?

adamruzicka:
Yes. There are some edge cases where it is a bit different, but in general yes.

PR #67, opened 2020-04-16
Allow newer bundler versions

mmoll:


mmoll:
Fixes Jenkins on Ruby 2.7 with bundler 2.x. Ubuntu 20.04 will bring bundler 2, too.

adamruzicka:
Thank you @mmoll!

PR #68, opened 2020-04-26
Fixes #29622 - Drop webrick-patch

ekohl:
With current Ruby's webrick it's possible to set ciphers directly.


ezr-ondrej:
Why is the option not enough? We want to force the ciphers by ourselfs for some reason?

ekohl:
> Why is the option not enough? We want to force the ciphers by ourselfs for some reason?

It's not great, but this mirrors the Smart Proxy itself and previous behavior. Ruby out of the box allows some insecure ciphers. Keeping this patch as small as possible (but a bit too small now, updated).

adamruzicka:
Thank you @ekohl & @ezr-ondrej !

PR #72, opened 2020-08-28
Fixes #30732 - Use SdNotify library

adamruzicka:


ekohl (approved):
This is true for the other PRs as well, but does it make sense to pin to `~> 0.1`?

adamruzicka:
I don't expect sd_notify to change much, but added the pin to all 3 PRs, just in case

ekohl (approved):
:+1: pending packaging merge

lzap (approved):

PR #75, opened 2020-11-20
Fixes #31352 - separate journald bundler group

lzap:
We must not require logging-journald by default, users should be able to run without it in development or on production platforms which don't have journald. After this is separated, we need to add a metapackage similarly as smart proxy and also modify our installer to install it if journal logging is enabled (or if thats not configurable then always install it):

```
%package journald
Summary:  Foreman Proxy journald logging dependencies
Group:    Applications/System
Requires: %{?scl_prefix}rubygem(logging-journald) >= 2.0
Requires: %{?scl_prefix}rubygem(logging-journald) < 3.0
Requires: %{name} = %{version}-%{release}

%description journald
Additional dependencies required to configure journald logging.

%files journald
%{_datadir}/%{name}/bundler.d/journald.rb
```

### DIFF
--- a/bundler.d/journald.rb
+++ b/bundler.d/journald.rb
@@ -1,0 +1,6 @@
+# Disable to avoid journald native gem in development setup.
+# This group is only distributed in dynflow_core (not needed
+# for smart-proxy plugins as logging is provided by smart-proxy).
+group :journald do
+  gem 'logging-journald', '~> 2.0', :platforms => [:ruby]
+end
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,6 @@ group :test do
   gem 'rubocop', '~> 0.52.1'
 end
 
-gem 'logging-journald', '~> 2.0', :platforms => [:ruby]
 gem 'rack', '>= 1.1'
 gem 'sinatra'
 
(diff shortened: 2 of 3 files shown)


bundler.d/journald.rb
```diff
@@ -0,0 +1,6 @@
+# Disable to avoid journald native gem in development setup.
```
evgeni, line 1: shouldn't this be `bundler.plugins.d` (or the other file there also be in `bundler.d`)?

evgeni:
I don't think we need that file at all, just set it to `:require => false` and let the normal smart proxy journald integration care about installing it?

bundler.d/journald.rb
evgeni: nvm

adamruzicka:
>  Or does sp_dynflow_core not read those settings?

It reads only /etc/smart_proxy_dynflow_core/settings.yml

lzap:
> I don't think we need that file at all, just set it to `:require => false` and let the normal smart proxy journald integration care about installing it?

The plugin uses registration, when the gem gets loaded then it registers the implementation in the base classes. It needs to be required at some point somewhere.

Have we agreed on something?

lzap:
Replaced by https://github.com/theforeman/smart_proxy_dynflow/pull/76

PR #76, opened 2020-11-20
Fixes #31352 - don't require logging-journald

lzap:
Solves the issue mentioned in the ticket. The logging gem uses auto loading capability and discovers all plugins automatically, there is no need to require it.

However if uses configures journald he/she must ensure the gem was installed otherwise it will error out.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ group :test do
   gem 'rubocop', '~> 0.52.1'
 end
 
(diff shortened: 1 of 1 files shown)


evgeni (approved):
That's what I meant in the other PR!

lzap:
Reworded the commit and PR message a bit.

adamruzicka (approved):

adamruzicka:
Thank you @lzap & @evgeni !

PR #78, opened 2021-04-29
Merge smart_proxy_dynflow_core and smart_proxy_dynflow

adamruzicka:
`smart_proxy_dynflow_core` got reduced to a tiny compatibility layer for `*_core` plugins which depend on it directly. This can be dropped after all `*_core` plugins are merged with their smart proxy counterpart plugins.


settings.d/dynflow.yml.example
```diff
@@ -1,9 +1,3 @@
 ---
 :enabled: true
 :database: /var/lib/foreman-proxy/dynflow/dynflow.sqlite
-:core_url: 'http://127.0.0.1:8008'
```
ezr-ondrej, line 4: should we add `execution_plan_cleaner_age` and `console_auth` here?
Also, will the current settings get migrated somehow, or will that need to be handled manually by user?

ezr-ondrej (commented):
Looks and works great, just worried about the setting migration, but I guess we can do the migration in installer easily, so it's only manual installation will have problems?

In that case I believe we can just have an upgrade warning.

adamruzicka:
> but I guess we can do the migration in installer easily, so it's only manual installation will have problems?

The installer currently does not expose (or even set) `execution_plan_cleaner_age`. Would you be fine with exposing it in the installer and mentioning it in the upgrade notes?

ezr-ondrej:
Mentioning in upgrade notes is fine by me, setting it when we do not set it now is not neccessary IMHO

ezr-ondrej (approved):
I've put the upgrade notes to Action items document and retested, works nicely, thanks @adamruzicka  :+1:

PR #82, opened 2021-06-15
Fixes #32811 - Unify bundler*.d directories

adamruzicka:


ezr-ondrej (approved):
Thanks @adamruzicka ! :+1:

PR #83, opened 2021-06-15
Refs #32534 - add forgotten deps from dynflow_core

ezr-ondrej:
In e205e19a00d8ef269953811a3c4e5140776915b1 we've removed the dependency
on smart_proxy_dynflow_core, but we've forgotten to bring over the gem
runtime dependencies.

### DIFF
--- a/smart_proxy_dynflow.gemspec
+++ b/smart_proxy_dynflow.gemspec
@@ -21,7 +21,9 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '~> 2.5'
 
-  gem.add_runtime_dependency "logging"
+  gem.add_runtime_dependency('dynflow', "~> 1.1")
+  gem.add_runtime_dependency('rest-client')
+  gem.add_runtime_dependency('sqlite3')
 
   gem.add_development_dependency "bundler", ">= 1.7"
   gem.add_development_dependency 'minitest'


smart_proxy_dynflow.gemspec
```diff
@@ -21,7 +21,14 @@ Gem::Specification.new do |gem|
...
   gem.required_ruby_version = '~> 2.5'
 
-  gem.add_runtime_dependency "logging"
+  gem.add_runtime_dependency('dynflow', "~> 1.1")
+  gem.add_runtime_dependency('logging')
+  gem.add_runtime_dependency('rack')
+  gem.add_runtime_dependency('rest-client')
+  gem.add_runtime_dependency('sd_notify', '~> 0.1')
+  gem.add_runtime_dependency('sequel')
+  gem.add_runtime_dependency('sinatra')
```
ekohl, line 30: Is this really needed? Smart Proxy should take care of this

smart_proxy_dynflow.gemspec
```diff
@@ -21,7 +21,14 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '~> 2.5'
 
-  gem.add_runtime_dependency "logging"
+  gem.add_runtime_dependency('dynflow', "~> 1.1")
+  gem.add_runtime_dependency('logging')
+  gem.add_runtime_dependency('rack')
+  gem.add_runtime_dependency('rest-client')
+  gem.add_runtime_dependency('sd_notify', '~> 0.1')
```
ekohl, line 28: Is this really needed? Smart Proxy should take care of this

smart_proxy_dynflow.gemspec
```diff
@@ -21,7 +21,14 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '~> 2.5'
 
-  gem.add_runtime_dependency "logging"
+  gem.add_runtime_dependency('dynflow', "~> 1.1")
+  gem.add_runtime_dependency('logging')
+  gem.add_runtime_dependency('rack')
```
ekohl, line 26: Is this really needed? Smart Proxy should take care of this

smart_proxy_dynflow.gemspec
adamruzicka: This was only used when running smart proxy dynflow core as a standalone service, also it is brought in by smart proxy itself

smart_proxy_dynflow.gemspec
adamruzicka: Those are brought in by smart proxy itself

smart_proxy_dynflow.gemspec
```diff
@@ -21,7 +21,10 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '~> 2.5'
 
-  gem.add_runtime_dependency "logging"
+  gem.add_runtime_dependency('dynflow', "~> 1.1")
+  gem.add_runtime_dependency('rest-client')
+  gem.add_runtime_dependency('sequel')
```
adamruzicka, line 26: This should be brought in by dynflow, iirc we don't use it directly anywhere else

adamruzicka (approved):
APT

ezr-ondrej:
I think, we could even do without the RestClient as it is only required for one use, but this is easier than refactoring.

adamruzicka:
Thank you @ezr-ondrej & @ekohl !

smart_proxy_dynflow.gemspec
```diff
@@ -21,7 +21,9 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '~> 2.5'
 
-  gem.add_runtime_dependency "logging"
+  gem.add_runtime_dependency('dynflow', "~> 1.1")
+  gem.add_runtime_dependency('rest-client')
```
ekohl, line 25: Just a random note: in smart-proxy itself there's also multiple similar implementations for HTTP abstraction. I think it makes sense to consider a common solution. Food for thought.

PR #84, opened 2021-06-18
Fix OutputCollector superclass

adamruzicka:


ares (approved):
Helps with the problem :+1:

ares:
Thanks @adamruzicka, merging!

PR #86, opened 2021-06-23
Refs #32534 - dont use SmartProxyDynflowCore

ezr-ondrej:
Remove usage of SmartProxyDynflowCore.

### DIFF
--- a/lib/smart_proxy_dynflow/action/shareable.rb
+++ b/lib/smart_proxy_dynflow/action/shareable.rb
@@ -12,14 +12,14 @@ module Proxy::Dynflow::Action
       planned_action = plan_self(input)
       # code only applicable, when run with SmartProxyDynflowCore in place
       if on_proxy? && callback
-        plan_action(SmartProxyDynflowCore::Callback::Action, callback, planned_action.output)
+        plan_action(Proxy::Dynflow::Callback::Action, callback, planned_action.output)
       end
     end
 
     private
 
     def on_proxy?
-      defined?(SmartProxyDynflowCore::Callback)
+      true
     end
   end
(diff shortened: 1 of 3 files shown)


lib/smart_proxy_dynflow/action/shareable.rb
```diff
@@ -12,14 +12,14 @@ def plan(input)
...
+        plan_action(Proxy::Dynflow::Callback::Action, callback, planned_action.output)
       end
     end
 
     private
 
     def on_proxy?
-      defined?(SmartProxyDynflowCore::Callback)
+      defined?(Proxy::Dynflow::Callback)
     end
```
adamruzicka, line 23: This should now always be true, the action cannot run anywhere else than on the smart proxy

ezr-ondrej: Do we want to remove that then and remove usage of it?
Or just return always `true` and have a followup?

ezr-ondrej: I feel the latter is what we want, so I went with it, I'll try to come up with follow-up

adamruzicka: It doesn't seem to be used outside of this repository so I'd be fine with removing it along with its usage in this pr

adamruzicka: Hah, github didn't refresh. Nevermind then, let's do it in a followup

adamruzicka (approved):
Works well, ACK

adamruzicka:
Thank you @ezr-ondrej !

PR #87, opened 2021-06-23
Drop smart_proxy_dynflow_core deps, add constraint on sp_dynflow

adamruzicka:


smart_proxy_dynflow_core.gemspec
```diff
@@ -22,20 +22,5 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = '~> 2.5'
 
   gem.add_development_dependency "bundler", ">= 1.7"
```
adamruzicka, line 24: Not sure if we even need this

ezr-ondrej: now we don't given we don't allow this deployed separatelly, `smart-proxy` should take care of that.

ezr-ondrej (approved):
IMHO we want to drop the bundler dependency as well, as it doesn't make much sense for a wrapper gem, but it's very open and development only, so lets get this in without it :+1:

PR #88, opened 2021-06-24
Update readme

adamruzicka:


README.md
```diff
@@ -57,11 +57,14 @@ END
...
 ```
 
+Note: The example above requires `smart_proxy_remote_execution_ssh` Smart Proxy
+plugin.
+
 #### POST /tasks/$TASK_ID/cancel
 Tries to cancel a task.
 
 ```
-curl -X POST localhost:8008/tasks/dd5a8306-0e52-4f68-9e83-c7f51c9e95c3/cancel -d '' 2>/dev/null | jq
```
ezr-ondrej, line 64: Why we dont like `j1` nomore?

ezr-ondrej (approved):
:+1: I don't get removing `jq`, but that's a detail, let's get this in :)

README.md
adamruzicka: It is not really essential to what one might want to accomplish with this and I didn't want to force my preferred tools on anyone

PR #90, opened 2021-07-08
Fixes #32961 - Groundwork for async batch triggering

adamruzicka:


adamruzicka:
Does it make sense to have separate sync and async variants or should we just switch from one to the other once it is available?

adamruzicka:
[test smart_proxy_dynflow]

adamruzicka:
:green_apple:

ezr-ondrej (approved):
:shipit:

PR #91, opened 2021-07-09
Fixes #32970 - Use ForemanRequest when calling back to Foreman

adamruzicka:


ekohl (approved):
Thanks! Yay code reuse.

PR #92, opened 2021-07-12
Completely drop smart_proxy_dynflow_core

adamruzicka:


ezr-ondrej (approved):
This got missed by me. We should get rid of this.
Hopefully we don't break nightlies again, but lets hope fore the best :+1:

PR #93, opened 2021-07-21
Refs #32024 - Register a launcher mimicing the old API

adamruzicka:


ezr-ondrej (approved):
Unsure about the failures, but they are definitelly not caused by this, so lets hope here as it looks safe to me :)
Thanks @adamruzicka !

PR #100, opened 2021-11-16
Fixes #33932 - allow cancellation of Group tasks

ezr-ondrej:
Group tasks have only one Runner per Group.
This disable the ability to cancel single tasks.
It is only possible to cancel the whole batch, but we were not allowing even that.

To note: this will cancel the whole group, not just the single task obviously.
Is that something we want, or do we need the complete solution to indicate this on foreman side?

@adamruzicka opinions?

### DIFF
--- a/lib/smart_proxy_dynflow/action/output_collector.rb
+++ b/lib/smart_proxy_dynflow/action/output_collector.rb
@@ -4,5 +4,10 @@ module Proxy::Dynflow::Action
       output[:result] = []
       suspend
     end
+
+    def kill_run
+      execution_plan = world.persistence.load_execution_plan(caller_execution_plan_id)
+      execution_plan.cancel
+    end
   end
 end
--- a/lib/smart_proxy_dynflow/action/single_runner_batch.rb
+++ b/lib/smart_proxy_dynflow/action/single_runner_batch.rb
@@ -1,7 +1,10 @@
 module Proxy::Dynflow::Action
   class SingleRunnerBatch < Batch
+    include ::Dynflow::Action::Cancellable
+
     def plan(launcher, input_hash)
       launcher.launch_children(self, input_hash)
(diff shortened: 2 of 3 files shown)


adamruzicka:
To me as someone who knows the rationale why, it seems fine. However, without knowing the inner workings the behavior might appear random. From foreman's side, hosts get batched by some non-apparent criterion and then when you cancel the run for a single host you won't really know which hosts it might take down with it.

ezr-ondrej:
Agreed, so what do you think we should do here?
There are only two simple scenarios to implement
1) this - "Take out random hosts" but usually that's what you want as it you cancel the whole invocation
2) disable canceling for batches and proceed with IMHO proper fix

Proper fix: Keep information what hosts are in batch together and keep track what jobs allow only batch cancellation so foreman could allow only canceling the batch or all the invocation at once in the UI.

adamruzicka:
I'd be glad if we could do it properly. We will know which hosts are in a batch together once this gets in https://github.com/theforeman/foreman-tasks/pull/641 . From there it should be fairly straightforward to implement, although a bit tricky to display

adamruzicka:
How about https://github.com/theforeman/smart_proxy_dynflow/pull/105 as an alternative?

adamruzicka:
This shouldn't now be needed as the issue should be already resolved by #105. Could you please verify it and then either close this or rebase?

ezr-ondrej:
We have the runner available now, but we still need to change how we deliver cancel event, right?

adamruzicka:
I don't think so. WIth https://github.com/theforeman/smart_proxy_dynflow/pull/105/files#diff-a7318c494ee78a035d048e365d66856cc309786346f576c23f22c1e6f5456882R5 the output collector action knows the id of the parent runner and it inherits cancel behavior from `Action::Runner` so it should just work.

PR #101, opened 2021-12-03
Fixes #34079 - Ensure output_result is an array

adamruzicka:


ezr-ondrej (approved):
Thanks @adamruzicka ! :+1:

PR #102, opened 2021-12-06
Fixes #34092 - Store even non-commited output chunks on finish

adamruzicka:
Output chunks are stored when the action is saved, but we we're aggregating them
into output[:result] before the final save so unsaved chunks didn't make it there.


ezr-ondrej (approved):
bit ugly, but works :shrug:

PR #103, opened 2022-02-03
Fixes #34376 - Provide convenient abstractions for local processes

adamruzicka:


ezr-ondrej:
I admit this feels much better then Command did. Tho should we deprecate `Command`?
And should we document usage of this abstraction layer a bit?

adamruzicka:
I'm trying to cover it with tests first, then I can put together some docs. Which brings a question, where should it be documented? Just in the readme?

ezr-ondrej:
> where should it be documented? Just in the readme?

I've started to put detailed docs in [`developer_docs`](https://github.com/theforeman/smart_proxy_dynflow/tree/782018979f07600d2562bce0f150f1102d17307d/developer_docs) and just point to it from README, but in general developer docs should be with the code :shrug: so documenting it in this repo make sense.

ezr-ondrej:
I dig the inline comments docs btw!

adamruzicka:
I put together some docs about runners in general. It doesn't cover all the stuff we do in sp-rex-ssh and sp-ansible, but I hope it is enough as a starting point.

PR #104, opened 2022-03-05
Update Rubocop & use reusable workflows

ekohl:
This started as an effort to use reusable workflows within the Foreman project. Then I realized that the GitHub formatter in Rubocop was unavailable, so this also updates the version to a more recent one. It still has some errors left so don't consider this ready at all. At this point I just want to see if the workflow works and annotations are created.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -11,6 +11,9 @@ concurrency:
   group: ${{ github.ref_name }}
   cancel-in-progress: true
 
+env:
+  BUNDLE_WITHOUT: development
+
 jobs:
   rubocop:
     name: Rubocop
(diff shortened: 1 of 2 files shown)


.github/workflows/ruby.yml
```diff
@@ -4,21 +4,14 @@ on:
...
-      - uses: actions/checkout@v2
-      - name: Setup Ruby
-        uses: ruby/setup-ruby@v1
-        with:
-          ruby-version: 2.5
-          bundler-cache: true
-      - name: Run rubocop
-        if: github.event_name != 'push'
-        run: bundle exec rubocop
+    uses: ekohl/actions/.github/workflows/rubocop.yml@v0
```
ekohl, line 13: It is my intention to move this to `theforeman` and `v1` before this should be merged.

ekohl:
#122 updated RuboCop, #128 did the reusable actions. I don't think the BUNDLE_WITHOUT part is working and https://github.com/theforeman/smart_proxy_dynflow/pull/130 does the RuboCop bit. Closing this.

PR #105, opened 2022-03-09
On demand

adamruzicka:


ezr-ondrej (commented):
:police_officer: is unhappy, I on the other hand like that it is quite minimal change for giving us quite a value.

adamruzicka:
Do we want to have the on demand processing enabled by default for all runners? For example for ssh we still need to consume data from the pipes, but we don't need to notify the action each time there is a bit of data.

ezr-ondrej:
I'd say it make sense to have them all on demand, I guess waking the actions all the time can consume quite a lot of resources...

lib/smart_proxy_dynflow/action/runner.rb
```diff
@@ -14,6 +16,8 @@ def run(event = nil)
           process_external_event(event)
         when ::Dynflow::Action::Cancellable::Cancel
           kill_run
+        when ::Proxy::Dynflow::Action::WithExternalPolling::Poll
+          poll
```
ezr-ondrej, line 20: we have run defined in the `WithExternalPolling` module don't we want to deffer this to the module?
Our poll is suspending, but the module's `run` calls poll and then suspending, so it might cause confusion and doesn't look too DRY

ezr-ondrej (commented):
I'm tiny bit confused, too many `refresh` named methods, can we maybe meet and document (in some UML e.g.) how the refresh/fetch of output is handled?

lib/smart_proxy_dynflow/action/runner.rb
adamruzicka: Made those match, module `#poll` does nothing, class `#poll` just polls. In both, `#poll` gets called in `#run` and then the action suspends itself

ezr-ondrej (approved):
Looks reasonable and works, thanks @adamruzicka  :+1:

PR #106, opened 2022-03-15
Fixes #34616 - Create DB directory dynamically

ekohl:
If the DB file is /var/lib/foreman-proxy/dynflow/dynflow.sqlite and /var/lib/foreman-proxy/dynflow doesn't exist, it fails. This ensures the directory exists. In practice foreman-proxy already creates /var/lib/foreman-proxy so this should work.

I haven't had a chance to verify this. @janniswarnat do you have a chance to verify this solves your problem?

Another solution is to keep the DB in memory which can be achieved by leaving the setting empty. This is what the foreman-installer does by default.

### DIFF
--- a/lib/smart_proxy_dynflow/core.rb
+++ b/lib/smart_proxy_dynflow/core.rb
@@ -1,3 +1,5 @@
+require 'fileutils'
+
 module Proxy::Dynflow
   class Core
     attr_accessor :world, :accepted_cert_serial
@@ -29,6 +31,7 @@ module Proxy::Dynflow
         Log.instance.warn "Could not open DB for dynflow at '#{db_file}', " \
                           "will keep data in memory. Restart will drop all dynflow data."
       else
+        FileUtils.mkdir_p(File.dirname(db_file))
         db_conn_string += "/#{db_file}"
       end
 


adamruzicka:
Could we just default to in-memory db? On-disk sqlite is terrible when concurrent writes are involved

ekohl:
Are you suggesting to change this to an empty value?
https://github.com/theforeman/smart_proxy_dynflow/blob/6ee4ad0d5545f82a9fa27bf2a886a30df1b82da4/settings.d/dynflow.yml.example#L3
I think that would change the default of the packaging.

adamruzicka:
> Are you suggesting to change this to an empty value?

Yes

>  I think that would change the default of the packaging.

Is that a problem, given the default isn't the sanest of options?

ekohl:
> Is that a problem, given the default isn't the sanest of options?

No, not a problem. I was just stating the implications, which would solve the bug. Another (positive) implication is that the diff that the installer applies is smaller. Something I always think is a good thing.

janniswarnat:
> If the DB file is /var/lib/foreman-proxy/dynflow/dynflow.sqlite and /var/lib/foreman-proxy/dynflow doesn't exist, it fails. This ensures the directory exists. In practice foreman-proxy already creates /var/lib/foreman-proxy so this should work.
> 
> I haven't had a chance to verify this. @janniswarnat do you have a chance to verify this solves your problem?
> 
> Another solution is to keep the DB in memory which can be achieved by leaving the setting empty. This is what the foreman-installer does by default.

@ekohl This in fact fixes the issue. If I mount your adapted core.rb file into my container, the foreman-proxy starts up just fine and the file `/var/lib/foreman-proxy/dynflow/dynflow.sqlite` is being created.

ekohl:
@adamruzicka I'd say we should implement both then. This in itself is a good fix, but you're right that defaulting to pure in-memory is also good. Thoughts?

adamruzicka:
Best of both worlds, what's not to like? Will there be a followup PR or will both changes happen here?

ekohl:
I'll do a follow up for the new default.

adamruzicka (approved):

adamruzicka:
Thank you @ekohl !

PR #107, opened 2022-03-17
Fixes #34625 - Make execution plan cleaner more aggressive

adamruzicka:
It will now wake up every half an hour and remove all execution plans
which have been stopped for at least half an hour. Half an hour should
give Foreman enough time to check up on the proxy in case the callback
from proxy fails, while not being too high at the same time.

### DIFF
--- a/lib/smart_proxy_dynflow/plugin.rb
+++ b/lib/smart_proxy_dynflow/plugin.rb
@@ -11,7 +11,7 @@ module Proxy::Dynflow
     settings_file "dynflow.yml"
     requires :foreman_proxy, ">= 1.16.0"
     default_settings :console_auth => true,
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
Very aggresive change from 24 hours to half an hour. I think it should be fine tho.
Should we have an upgrade warning about it tho?
There might be ppl relying on this for some reason, so just in case :shrug:

adamruzicka:
Favourite question: Where should the upgrade warning live?

ezr-ondrej:
eh, right we do not have that here, I believe it should be living alongside the core warnings, but given it was not yet settled, I'd be totally up for spinning it up in plugins documentations, but that is too much work, so just not add ourselves the headache given it's not gonna affect too many ppl?

PR #109, opened 2022-03-28
Fixes #34688 - Handle missing executable more gracefully

adamruzicka:
in ProcessManager and friends

This should also fix https://projects.theforeman.org/issues/31888 without us having to do any further changes to sp-ansible

### DIFF
--- a/lib/smart_proxy_dynflow/process_manager.rb
+++ b/lib/smart_proxy_dynflow/process_manager.rb
@@ -76,12 +76,12 @@ module Proxy
         out_read, out_write = IO.pipe
         err_read, err_write = IO.pipe
 
-        @pid = spawn(*@command, :in => in_read, :out => out_write, :err => err_write)
-        [in_read, out_write, err_write].each(&:close)
-
         @stdin.io  = in_write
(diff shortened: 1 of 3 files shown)


adamruzicka:
Not sure why travis is complainig, let's try once more
[test smart_proxy_dynflow]

lib/smart_proxy_dynflow/process_manager.rb
```diff
@@ -158,8 +158,10 @@ def close
       # @return [void]
       def finish
         close
-        _pid, status = Process.wait2(@pid)
-        @status = status.exitstatus
+        unless @status
```
ezr-ondrej, line 161: unless `pid == -1` ? :)

ezr-ondrej (approved):
One nit, but looks definitelly lovely to have ! :+1:

lib/smart_proxy_dynflow/process_manager.rb
adamruzicka: Right

adamruzicka:
Come on, mr. J
[test smart_proxy_dynflow]

PR #112, opened 2022-07-12
Fixes #35202 - Add API to expose details of execution plans

adamruzicka:
for multiple execution plans in a single request


adamruzicka:
This is just a new endpoint, it can't really hurt anything. Let's merge this

PR #113, opened 2022-09-14
Fixes #35526 - Make ProcessManager#finish reentrant

adamruzicka:


adamruzicka:
Per https://community.theforeman.org/t/remote-action-run-ansible-roles-stuck-pending-24h/30331/19?u=aruzicka this seems to work

adamruzicka (commented):
It's been a while, community member confirmed this works, let's go with it

PR #114, opened 2022-09-20
Fixes #35540 - Properly assemble callback url

adamruzicka:


ianballou (approved):
Tested with https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/95

PR #116, opened 2023-03-07
Upgrade to Ruby 3.1

archanaserver:
This PR include following changes:
- [ ]  #118 
- [ ]  #119 
- [ ]  #120


.rubocop.yml
```diff
@@ -1,3 +1,4 @@
+require: rubocop-rails
```
adamruzicka, line 1: Why? There's no rails in the project

.ruby-version
```diff
@@ -0,0 +1 @@
+3.1.2
```
adamruzicka, line 1: Is this necessary? People should be free to use whatever they please

adamruzicka (commented):
Does this do anything else than bumping rubocop and fixing the breakage that follows?

.rubocop.yml
archanaserver: Yes we don't need this, but can you help why we have this in here: https://github.com/theforeman/smart_proxy_dynflow/blob/master/.rubocop.yml#L6-L7. I guess we can drop it.

.ruby-version
archanaserver: No this is not necessary this get commited by mistake. I'll drop it, in the next commit.

.rubocop.yml
adamruzicka: My guess would be the version we had previously was carrying rails-specific cops which were controlled by what you linked, but they have been moved into rubocop-rails since then.

No matter why, I'd remove this and the rails.enabled: false thing and we'll see what happens

ekohl: My approach is to use blame, which points to https://github.com/theforeman/smart_proxy_dynflow/commit/a4740fccacfcfdc92db5000d4e4aa9c777c1950b. As @adamruzicka suggests, predates the extraction of those cops to rubocop-rails. If you look at https://github.com/rubocop/rubocop-rails/blob/master/CHANGELOG.md#200-2019-05-22 then version 2.0.0 from 2019-05-22 is newer than that commit and has this line:
> Extract Rails cops from rubocop/rubocop repository

archanaserver: @ekohl can you help me with this? it's the first time i find about blame.

archanaserver: Thanks @ekohl!

lib/smart_proxy_dynflow/io_buffer.rb
```diff
@@ -71,7 +71,7 @@ def close
       def read_available!
         data = ''
         loop { data += @io.read_nonblock(4096) }
-      rescue IO::WaitReadable # rubocop:disable Lint/HandleExceptions
```
ekohl, line 74: This was changed to `Lint/SuppressedException`, which is why it now shows up in the TODO file

Gemfile
```diff
@@ -14,7 +14,8 @@ group :test do
   gem 'public_suffix'
   gem 'rack-test'
   gem 'rake'
-  gem 'rubocop', '~> 0.52.1'
+  gem 'rubocop', '~> 1.28.0'
+  # gem 'rubocop-rails', require: false
```
ekohl, line 18: Generally speaking we avoid committing commented code
```suggestion
```

ekohl:
[test smart_proxy_dynflow]

ekohl:
This plugin is still using the old PR testing in Jenkins. There are GitHub Actions, but they don't run Ruby 3 either.

ekohl:
[test smart_proxy_dynflow]

ekohl (commented):
It now fails because there's no Jenkins task defined. See https://github.com/theforeman/smart_proxy_discovery/blob/bf92993f81037e3dd0ec47d61cd654c07d6475bd/Rakefile#L22-L25 for an example.

archanaserver:
> It now fails because there's no Jenkins task defined. See https://github.com/theforeman/smart_proxy_discovery/blob/bf92993f81037e3dd0ec47d61cd654c07d6475bd/Rakefile#L22-L25 for an example.

@ekohl i've defined the `jenkins` task, does it need any other changes?

ekohl:
I don't see the task and you can see the tests fail in `smart_proxy_dynflow` with:
```
Using /usr/local/rvm/gems/ruby-2.7.4 with gemset smart_proxy_dynflow-pr-test-1
rake aborted!
Don't know how to build task 'jenkins:unit' (See the list of available tasks with `rake --tasks`)
```

So you need to define `jenkins:unit` for it to run.

archanaserver:
> 



> I don't see the task and you can see the tests fail in `smart_proxy_dynflow` with:
> 
> ```
> Using /usr/local/rvm/gems/ruby-2.7.4 with gemset smart_proxy_dynflow-pr-test-1
> rake aborted!
> Don't know how to build task 'jenkins:unit' (See the list of available tasks with `rake --tasks`)
> ```
> 
> So you need to define `jenkins:unit` for it to run.

@ekohl How can I do this? Can you help me with any resources to do the same?

archanaserver:
@adam I started facing an error in the pipeline that the `xml` files are not getting generated. But I can see the tests are running fine, also I tried running it locally but can't see those `xml` files getting generated even locally. So are we using different framework for tests here? Can you take a look at once here?

```
jenkins/reports/unit/*.xml— Archive JUnit-formatted test results<1s
Recording test results
No test report files were found. Configuration error?
```

https://ci.theforeman.org/blue/organizations/jenkins/smart_proxy_dynflow-pr-test/detail/smart_proxy_dynflow-pr-test/5/

.rubocop.yml
```diff
@@ -3,9 +3,6 @@ inherit_from: .rubocop_todo.yml
 AllCops:
   TargetRubyVersion: 2.5
```
ekohl, line 4: Please update this to 2.7 and also enable new cops + additional plugins. It should look like:

```yaml
require:
  - rubocop-minitest
  - rubocop-performance
  - rubocop-rake

AllCops:
  NewCops: enable
  TargetRubyVersion: '2.7'
```

Note the additional plugins need to be included as a dependency in `Gemfile`.

ekohl (changes requested):
Looking at this PR, I think it's getting fairly large and it's combining multiple things. I'd urge you to split the effort. I always link to https://theforeman.org/2019/03/merge-time-review-complexity.html but while searching for it I also found https://athenian.com/blog/why-pull-request-size-matters.

The tricky thing is learning to recognize where to set the boundary and that's something you learn over time. In this PR there are 3 things ongoing:

1. Update RuboCop
2. Make sure tests pass in Jenkins
3. Validate Ruby 3.1 compatibility

Step 2 depends on step 1, so I would first focus on that and that alone. It's OK if Jenkins fails, as long as RuboCop is updated. This repository already has tests using GitHub Actions so that's sufficient to test.

As for step 2, the explanation is that this repository uses [minitest](https://github.com/minitest/minitest) and not [test-unit](https://test-unit.github.io/). This means you should use [minitest-reporters](https://github.com/minitest-reporters/minitest-reporters) instead of [ci_reporter_test_unit](https://github.com/ci-reporter/ci_reporter_test_unit). https://github.com/theforeman/foreman_maintain/pull/711 is a good example of how integration happens. Today Jenkins already runs tests on Ruby 3.1, but it would be OK to keep that failing and only make sure Ruby 2.7 passes.

Then step 3 is one you can probably already answer after step 2. Perhaps it's not needed at all and it "just works".

Depending on how large step one is, you can even decide to split it. Often you can already fix some cops before updating RuboCop itself. Separating that into multiple commits or even multiple pull requests can be friendly on the reviewer.

A common thing is that you use a command to make changes. If you do, put the command in the commit message so the reviewer knows it. An example of that is `rubocop --autocorrect --only The/Cop`, but `sed` is another.

And answering @adamruzicka's question: from the platform team we're working on Ruby 3(.1) compatibility in all Smart Proxy plugins. We've chosen to utilize Jenkins because it provides a common interface and a single place to maintain the list of supported Ruby versions. An alternative would have been to use https://github.com/theforeman/actions but most Smart Proxy plugins are tested by Jenkins already so this was a smaller step.

A bit reason to use RuboCop is that various cops can replace deprecated methods that will break in Ruby 3.2, so it's also looking forward.

archanaserver:
Thanks @ekohl for the explaination and clearing all my doubts.
I'll make sure of it from now on, while starting to work on any PR and so. Also Thanks a lot for sharing those resources.

.rubocop.yml
archanaserver: @ekohl okay one more confusion i've, so when I added `rubocop-minitest` and rest of the `gems`.  I can see around 150 offenses created and mostly get generated from `minitest` only. I want to know, How I'm suppose to fix it? Using `auto-correct` isn't always right is it or there is some specific cases to use it? Because if I go and check manually it'll take a lot more time.

lib/smart_proxy_dynflow/runner/base.rb
```diff
@@ -62,7 +62,7 @@ def publish_data(data, type)
         @continuous_output.add_output(data, type)
       end
 
-      def publish_exception(context, exception, fatal: true)
+      def publish_exception(context, exception, fatal = true)
```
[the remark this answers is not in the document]
archanaserver: I guess `remote_execution_ssh` still use that.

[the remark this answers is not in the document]
adamruzicka: Oh, I see. You flipped it to the keyword form in c38dafe and then back in fe698dd . Nevermind then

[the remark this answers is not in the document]
archanaserver: Yes.

.rubocop.yml
```diff
@@ -35,8 +32,8 @@ Naming/FileName:
 Naming/HeredocDelimiterNaming:
   Enabled: false
 
-Style/TrailingCommaInLiteral:
-  Enabled: false
+# Style/TrailingCommaInLiteral:
```
ekohl, line 35: Please drop the commented lines

lib/smart_proxy_dynflow/io_buffer.rb
```diff
@@ -88,7 +88,7 @@ def write_available!
           n = @io.write_nonblock(@buffer)
           @buffer = @buffer[n..-1]
         end
-      rescue IO::WaitWritable # rubocop:disable Lint/HandleExceptions
+      rescue IO::WaitWritable
```
ekohl, line 91:
```suggestion
      rescue IO::WaitWritable # rubocop:disable Lint/SupressedException
```

lib/smart_proxy_dynflow/io_buffer.rb
```diff
@@ -71,7 +71,7 @@ def close
       def read_available!
         data = ''
         loop { data += @io.read_nonblock(4096) }
-      rescue IO::WaitReadable # rubocop:disable Lint/HandleExceptions
+      rescue IO::WaitReadable
```
ekohl, line 74:
```suggestion
      rescue IO::WaitWritable # rubocop:disable Lint/SupressedException
```

ekohl (changes requested):

.rubocop.yml
archanaserver: Updated this changes here https://github.com/theforeman/smart_proxy_dynflow/pull/118

I have moved Rubocop and jenkins work to #118.

lib/smart_proxy_dynflow/io_buffer.rb
archanaserver: Updated this changes here https://github.com/theforeman/smart_proxy_dynflow/pull/118

lib/smart_proxy_dynflow/io_buffer.rb
archanaserver: Updated this changes here https://github.com/theforeman/smart_proxy_dynflow/pull/118

archanaserver:
In favour of this PR and making it more clean and easy for review. I've created a new PR -> #121 and I'm working on it.

Thanks for the review and pointing me out through your suggestions @ekohl.
Apologies for making it a mess, I'm still learning and will make sure to make it easy for your review :)

PR #123, opened 2023-07-10
Drop Ruby 2.5 support

archanaserver:
For refernce: https://github.com/theforeman/smart_proxy_dynflow/pull/122#issuecomment-1625472248

PR #124, opened 2023-07-24
Drop smart_proxy_dynflow Jenkins test changes

archanaserver:
For refernce: https://github.com/theforeman/smart_proxy_dynflow/pull/123#issuecomment-1646601180

We have GitHub Action to run the tests against different Ruby versions.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,6 @@ end
 
 group :test do
   gem 'minitest'
-  gem 'minitest-reporters'
   gem 'mocha'
   gem 'public_suffix'
   gem 'rack-test'
--- a/Rakefile
+++ b/Rakefile
@@ -29,8 +29,3 @@ if defined? RuboCop
     task.fail_on_error = true
   end
 end
-
-namespace :jenkins do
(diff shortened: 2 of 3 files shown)


archanaserver:
@ehelms does this looks good?

Gemfile
```diff
@@ -16,7 +16,6 @@ end
 
 group :test do
   gem 'minitest'
-  gem 'minitest-reporters'
```
ehelms, line 19: I think keeping this is good, as these reporters are useful for more than just Jenkins. @ekohl ?

archanaserver: Should I keep it now @ehelms @ekohl? I don't see any other configuration for it to use, other than Jenkins in this project.

ekohl: It slightly changes the output and gives a progress bar but I'm not sure I ever found that very useful. I'm fine with either way.

ekohl (approved):
@adamruzicka any preference on minitest-reporters?

adamruzicka:
I wouldn't mind if they went away.

PR #131, opened 2024-01-18
Fixes #37060 - Distinguish between bytes and chars

adamruzicka:


pondrejk (approved):
checked on patched stream, non-ascii pwd no longer causes a problem for the effective user. Effective user related tests from robottelo also pass

PR #132, opened 2024-01-19
Fix flaky tests

adamruzicka:


test/process_manager_test.rb
```diff
@@ -91,6 +91,7 @@ class ProcessManagerTest
           end
           pm.stdin.add_data("10\n")
           pm.run!
+          assert_equal '', pm.stderr.to_s
```
adamruzicka, line 94: This doesn't necessarily fix anything, but asserting there were no errors could help when this thing actually fails.

ofedoren (approved):
Thanks, @adamruzicka, changes look reasonable, tests are green, let's get this in.

PR #134, opened 2024-03-20
Fixes #37284 - Do not overwrite results

adamruzicka:


lib/smart_proxy_dynflow/action/middleware/assemble_results.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Proxy::Dynflow::Action::Middleware
+  class AssembleResults < ::Dynflow::Middleware
+    def present
+      unless %i[error success skipped].include?(action.run_step&.state)
+        action.output[:result] = action.output_result
+      end
```
ofedoren, line 8: If I read it right, we only add `result` entry for stopped/paused states since otherwise it's already present?

ofedoren (approved):
Thanks, @adamruzicka, LGTM.

lib/smart_proxy_dynflow/action/middleware/assemble_results.rb
adamruzicka: Yes, it should already be there from https://github.com/theforeman/smart_proxy_dynflow/pull/134/files#diff-f3c605b1717c7b0b34ac0715dcb3867c32689fad5a09c7f73a9de47708fa326bR66

PR #135, opened 2024-04-05
Fixes #37319 - Allow setting IDs for outputs

adamruzicka:
and switch runner output interface to be keyword based.

### DIFF
--- a/lib/smart_proxy_dynflow/continuous_output.rb
+++ b/lib/smart_proxy_dynflow/continuous_output.rb
@@ -37,18 +37,20 @@ module Proxy::Dynflow
       raw_outputs.map { |output| output['output'] }.join("\n")
     end
 
-    def add_exception(context, exception, timestamp = Time.now.getlocal)
-      add_output(context + ": #{exception.class} - #{exception.message}", 'debug', timestamp)
+    def add_exception(context, exception, timestamp: Time.now.getlocal, id: nil)
+      add_output(context + ": #{exception.class} - #{exception.message}", 'debug', timestamp: timestamp, id: id)
     end
 
-    def add_output(*args)
-      add_raw_output(self.class.format_output(*args))
+    def add_output(...)
+      add_raw_output(self.class.format_output(...))
     end
 
-    def self.format_output(message, type = 'debug', timestamp = Time.now.getlocal)
-      { 'output_type' => type,
-        'output' => message,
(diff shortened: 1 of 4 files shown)


lib/smart_proxy_dynflow/continuous_output.rb
```diff
@@ -37,18 +37,20 @@ def humanize
       raw_outputs.map { |output| output['output'] }.join("\n")
     end
 
-    def add_exception(context, exception, timestamp = Time.now.getlocal)
-      add_output(context + ": #{exception.class} - #{exception.message}", 'debug', timestamp)
+    def add_exception(context, exception, timestamp: Time.now.getlocal, id: nil)
```
ekohl, line 40: Isn't this an API change?
```ruby
def add_exception(context, exception, timestamp = 'now')
  puts timestamp
end

def add_exception_new(context, exception, timestamp: 'now')
  puts timestamp
end

add_exception('a', 'b')
add_exception('a', 'b', 'explicit')

add_exception_new('a', 'b')
add_exception_new('a', 'b', 'explicit')
```

```console
$ ruby test2.rb
now
explicit
now
test2.rb:5:in `add_exception_new': wrong number of arguments (given 3, expected 2) (ArgumentError)
	from test2.rb:13:in `<main>'
```

Or do you accept it since it only appears to be used internally.

adamruzicka: While this isn't strictly private api, none of the smart proxy plugins which depend on sp-dynflow seem to use this, so I'd be willing to let it slide.

lib/smart_proxy_dynflow/continuous_output.rb
```diff
@@ -37,18 +37,20 @@ def humanize
...
+      add_output(context + ": #{exception.class} - #{exception.message}", 'debug', timestamp: timestamp, id: id)
     end
 
-    def add_output(*args)
-      add_raw_output(self.class.format_output(*args))
+    def add_output(*args, **kwargs)
+      add_raw_output(self.class.format_output(*args, **kwargs))
```
ekohl, line 45: Would it make sense to also just forward the arguments here?
```suggestion
    def add_output(...)
      add_raw_output(self.class.format_output(...))
```

ekohl (approved):

PR #141, opened 2026-01-05
Allow dynflow-2.y.z

adamruzicka:


pondrejk (approved):

ofedoren (approved):
Thanks, @adamruzicka !

lib/smart_proxy_dynflow/api.rb
```diff
@@ -13,7 +13,7 @@ class Api < ::Sinatra::Base
 
       include ::Sinatra::Authorization::Helpers
 
-      TASK_UPDATE_REGEXP_PATH = %r{/tasks/(\S+)/(update|done)}.freeze
+      TASK_UPDATE_REGEXP_PATH = %r{/tasks/(\S+)/(update|done)}
```
ofedoren, line 16: Just to make sure: is this due to Ruby 3.0 Regex objects should be frozen by default or it triggered an error?

adamruzicka: Apparently it is an unnecessary noop
```
lib/smart_proxy_dynflow/api.rb:16:33: C: [Correctable] Style/RedundantFreeze: Do not freeze immutable objects, as freezing them has no effect.
      TASK_UPDATE_REGEXP_PATH = %r{/tasks/(\S+)/(update|done)}.freeze
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

